PR #7, opened 2024-10-01
🛠️ Feat/1 Setup Repo

demariadaniel:
- Creates basic React UI, Express API, and Types package, using Vite for set up
- Set up unit testing framework using Node Test Runner
- Adds basic README docs


apps/api/package.json
```diff
@@ -11,5 +11,9 @@
   "devDependencies": {
     "typescript": "^5.5.3",
     "vite": "^5.4.1"
+  },
+  "dependencies": {
+    "express": "^4.21.0",
+    "node": "^22.9.0"
```
justincorrigible, line 17: think you mean this one to be in "engines" instead, no?
not sure `npm i node` is a useful thing, but I'm very curious to see what this does!

demariadaniel: Yes (I'm curious too haha) but this PR is still in an extremely unfinished state so no need for feedback yet

packages/types/src/demoData.mts
```diff
@@ -0,0 +1,105 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { Application } from './types';
+
+export const demoApplication: Application = {
```
demariadaniel, line 22: Needed some demo data, so I created this based on early planning sessions. Expect this to completely change based on outcome of https://app.zenhub.com/workspaces/pcgl---pancanadian-genome-library-659eed8bc4fe0c04890cff77/issues/gh/pan-canadian-genome-library/daco/2

apps/api/src/main.mts
```diff
@@ -0,0 +1,47 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import cors from 'cors';
+import express, { Request, Response } from 'express';
+
+// TODO: fix import
+import { demoApplication } from 'pcgl-daco/packages/types/src/main.mjs';
```
ciaranschutte, line 24: is this just a typo? should be .mts instead of .mjs?

apps/ui/eslint.config.js
```diff
@@ -0,0 +1,28 @@
...
+    languageOptions: {
+      ecmaVersion: 2020,
+      globals: globals.browser,
+    },
+    plugins: {
+      'react-hooks': reactHooks,
+      'react-refresh': reactRefresh,
+    },
+    rules: {
+      ...reactHooks.configs.recommended.rules,
```
ciaranschutte, line 21: nothing wrong here, just a heads up - I think this will enforce far stricter hook rules than we're used to. specifically regarding hook dependencies

apps/ui/package.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "pcgl-daco-ui",
+	"private": true,
+	"version": "0.0.0",
+	"type": "module",
+	"scripts": {
+		"dev": "vite",
+		"build": "tsc -b && vite build",
```
ciaranschutte, line 8: I think you can just have tsc only typecheck (not build) and leave the building to Vite - afaik it works on ts files

packages/types/vite.config.ts
```diff
@@ -0,0 +1,17 @@
...
+// https://vitejs.dev/config/
+export default defineConfig({
+	build: {
+		// generate .vite/manifest.json in outDir
+		manifest: true,
+		lib: {
+			entry: './src/main.mts',
+			name: 'pcql-daco-types',
+		},
+		rollupOptions: {
```
ciaranschutte, line 12: curious why rollupOptions need to be defined with just an entry point. I think it's covered by lib already?

apps/ui/src/main.tsx
```diff
@@ -0,0 +1,10 @@
+import { StrictMode } from 'react'
+import { createRoot } from 'react-dom/client'
+import App from './App.tsx'
+import './index.css'
+
+createRoot(document.getElementById('root')!).render(
```
ciaranschutte, line 6: what does the `!` do here?

ciaranschutte:
looks good! I'd recommend adding prettier + husky at this point. standard in our stack. 
husky for pre commit checks and prettier for consistent styling
prettier-organise-imports package is useful too

apps/api/src/main.mts
```diff
@@ -0,0 +1,47 @@
...
+
+const app = express();
+
+// Exports only to demo test suite
+export const port = 3000;
```
ciaranschutte, line 29: at this setup point, likely good to add .env config - usually using the dotenv library - things like PORT should be configurable based on env.

ciaranschutte:
license could be added also at this point. kinda type that usually gets forgotten later!

package.json
```diff
@@ -0,0 +1,21 @@
...
+	"bugs": {
+		"url": "https://github.com/Pan-Canadian-Genome-Library/daco/issues"
+	},
+	"scripts": {
+		"test": "echo \"Error: no test specified\" && exit 1"
+	},
+	"keywords": [],
+	"author": "",
+	"license": "ISC"
+}
```
ciaranschutte, line 21: extra package.json?

joneubank: pnpm workspace root package, normal to have

apps/api/src/main.mts
joneubank: Good idea for a next PR is adding template for .env config.

apps/ui/src/main.tsx
joneubank: insists this will be not null. its a shortcut around TS type checking for either "I know better" or "Its ok for things to break here". This is becoming a pretty standard setup for React apps since if this node is missing the whole website will fail to render and so there isn't really a fallback.

apps/ui/eslint.config.js
joneubank: This seems like a good thing to start with, no?

apps/api/src/main.mts
joneubank: TS with EMS imports require .js file extension.

apps/api/src/main.mts
```diff
@@ -0,0 +1,47 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import cors from 'cors';
+import express, { Request, Response } from 'express';
+
+// TODO: fix import
```
joneubank, line 23: Whats this about? I wouldn't know how to action this todo.

apps/api/src/main.mts
```diff
@@ -0,0 +1,47 @@
...
+
+app.use(cors());
+
+app.get('/', (req: Request, res: Response) => {
+	console.log(req);
+	res.send('Hello World!');
+});
+
+app.get('/applications', cors(), (req: Request, res: Response) => {
+	// TODO: fix unused req
```
joneubank, line 39: Try `_req`. The underscore prefix indicates this is a placeholder parameter and is unused. TS will be happy with that. This TODO is why it may not make sense to enforce no unused params/locals.

apps/api/src/main.mts
```diff
@@ -0,0 +1,47 @@
...
+	console.log('Headers', req.headers);
+	res.send(demoApplication);
+});
+
+// Exports only to demo test suite
+export const server = app.listen(port, () => {
+	console.log(`Example app listening on port ${port}`);
+});
```
joneubank, line 47: Does `app.listen` create a new object, or could you just export `app` where it is declared and get the same effect? I think it would be clearer what is happening if you export the const where it is declared.

apps/api/src/tests/index.mjs
```diff
@@ -0,0 +1,68 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import assert from 'node:assert';
+import { beforeEach, describe, it } from 'node:test';
```
joneubank, line 21: Thank you for the `node:` prefix on node api imports.

apps/api/src/tests/index.mjs
```diff
@@ -0,0 +1,68 @@
...
+		});
+
+		await it('AsyncData does not equal 0', async () => {
+			assert.notEqual(asyncData, 0);
+		});
+
+		await it('AsyncData equals 2 (beforeEach called on every Test)', async () => {
+			assert.equal(asyncData, 2);
+		});
+	});
```
joneubank, line 56: Nice example usage. I think we can remove this from our setup though :)

apps/api/src/tests/index.mjs
```diff
@@ -0,0 +1,68 @@
...
+// Failure Example + Close Server
+describe('Third Test Suite', async () => {
+	it('Example Failure Case', async () => {
+		assert.strictEqual(port, 4000);
+	});
+
+	it('Close Server', async () => {
+		server.close();
+	});
+});
```
joneubank, line 68: Again, nice examples. Let's remove them from our setup.

apps/ui/README.md
```diff
@@ -0,0 +1,50 @@
+# React + TypeScript + Vite
```
joneubank, line 1: Lets replace this template readme with one related to the application we are actually providing. 

For inspiration you can look at the OHCRN repo for package readme examples:
https://github.com/OHCRN/platform/blob/develop/apps/consent-api/README.md
https://github.com/OHCRN/platform/blob/develop/apps/consent-ui/README.md

apps/ui/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<link rel="icon" type="image/png" href="/pcgl-logo.png" />
```
joneubank, line 5: nice!

packages/types/src/types.ts
```diff
@@ -0,0 +1,113 @@
...
+export enum Status {
+	'DRAFT',
+	'REVIEW',
+	'APPROVED',
+	'CLOSED',
+	'REVOKED',
+}
+
+export type Application = {
+	_type: 'Application';
```
joneubank, line 77: Type tagging? Looks like we're applying this inconsistently (used on revisions but not Project). Curious what the motivation for this is here.

joneubank (changes requested):
Really great starting point for this project, thanks. Left a handful of comments. In addition I want to suggest:

- Readmes:
   - no readme should say this is an Overture project. Its great to use some templating from other projects but we need to make sure the READMEs are all relevant to our project
   - Important to have a readme in each of the apps/packages, and we want these to be relevant from day 1. This means a relevant title, not a template description, and meaningful sections with configuration infomration, dev instructions, and any unique details a user or developer may be interested in
- TSConfig:
   - can you add some documentation explaining the split in tsconfigs between .app and .node? Or if this isnt needed, lets simplify the setup.
   - since we don't have scripts using one tsconfig vs the other, I'm pretty sure we only need a single tsconfig per project
   - It would be good to have a global tsconfig that all the projects reference. This would enforce standard linting rules (strict, unsused parameters, switch fallthrough, unchecked index, etc.) . Each package would then extend this base config and add their custom settings for bundler and target etc.
- Testing:
   - love seeing the node test demos, but we don't need these committed and left in our repo
- Prettier - Can we setup a global prettier config right from the start?

apps/ui/eslint.config.js
ciaranschutte: @joneubank 100%! as I say, nothing wrong at all, just a heads up. We had to tone down our eslint usage in a previous project (can't recall exactly which one) because we had so many things to fix. 

I agree - very good thing to start with - nice one @demariadaniel !

README.md
```diff
@@ -1 +1,67 @@
...
+| [Score](https://github.com/overture-stack/score/)       | Transfer data to and from any cloud-based storage system                                |
+| [Song](https://github.com/overture-stack/song/)         | Catalog and manage metadata associated to file data spread across cloud storage systems |
+| [Maestro](https://github.com/overture-stack/maestro/)   | Organizing your distributed data into a centralized Elasticsearch index                 |
+| [Arranger](https://github.com/overture-stack/arranger/) | A search API with reusable search UI components                                         |
+| [Stage](https://github.com/overture-stack/stage)        | A React-based front-data portal UI                                                      |
+| [Lyric](https://github.com/overture-stack/lyric)        | A data-agnostic tabular data submission system                                          |
+| [Lectern](https://github.com/overture-stack/lectern)    | A simple web browser UI that integrates Ego and Arranger                                |
```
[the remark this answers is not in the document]
demariadaniel: Re: Overture, I'm removing templated references, I was waiting for guidance on copy re: project summary and will review past meetings w/ Alexis

apps/ui/README.md
demariadaniel: Thanks, I need permissions to view

packages/types/src/types.ts
demariadaniel: Removed, this was more for noting how we should define our data as types, which is now implemented in the demo data

apps/api/src/main.mts
demariadaniel: Added specific comment. The goal is to get our export from the main package export, not from a specific file in `src`. So the `/packages` build needs some more configuring.

packages/types/package.json
```diff
@@ -0,0 +1,17 @@
+{
+	"name": "types",
```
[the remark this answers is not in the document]
demariadaniel: I've updated `types` to `data-model` following the example in Lyric: https://github.com/overture-stack/lyric/tree/main/packages
This better expresses what the package is used for but naming is hard so I'm open to other suggestions

README.md
```diff
@@ -1 +1,67 @@
...
+```
+
+The modules in the monorepo are organized into two categories:
+
+- **apps/** - Standalone processes meant to be run. These are published to [ghcr.io](https://ghcr.io) as container images.
+- **packages/** - Reusable packages shared between applications and other packages. Packages are published to [NPM](https://npmjs.com).
+
+## Local development
+
+### Development tools
```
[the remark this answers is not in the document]
demariadaniel: Updated

.gitignore
```diff
@@ -1,130 +1,15 @@
...
-
-# Yarn Integrity file
-.yarn-integrity
-
 # dotenv environment variable files
 .env
 .env.development.local
 .env.test.local
 .env.production.local
 .env.local
```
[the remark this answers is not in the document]
demariadaniel: Updated

apps/ui/tsconfig.app.json
```diff
@@ -0,0 +1,24 @@
...
+
+    /* Bundler mode */
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "isolatedModules": true,
+    "moduleDetection": "force",
+    "noEmit": true,
+    "jsx": "react-jsx",
+
+    /* Linting */
```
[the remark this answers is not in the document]
demariadaniel: Added

apps/api/.gitignore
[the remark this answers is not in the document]
demariadaniel: I was able to move everything to the root; the project level gitignores were all duplicates. I deleted the files; they can be added as needed.

docs/1-Overview.md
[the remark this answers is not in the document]
demariadaniel: All references removed

apps/api/src/main.mts
demariadaniel: Are you guys OK if I create a new issue to cover:
- Add env support
- Add Husky
- Add Prettier
- Add License
then add those in a follow up PR?

apps/api/src/main.mts
demariadaniel: `.listen` does return a `server` which is different from `app` (returned from `express()`)
So I wouldn't be able to export `app` and call `.close`

See:
- https://expressjs.com/en/advanced/healthcheck-graceful-shutdown.html
- https://expressjs.com/en/5x/api.html#app.listen

apps/api/src/tests/index.mjs
demariadaniel: Removed. Kept only the first test case.

apps/ui/tsconfig.app.json
```diff
@@ -0,0 +1,24 @@
...
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
```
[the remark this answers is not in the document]
demariadaniel: Yeah I think linter is appropriate. It's always good to remove unused imports but with variables it can get tricky when there are unused positional arguments. The Express `req` example is the perfect example.

apps/api/src/main.mts
demariadaniel: Thank you did not know about that trick

apps/ui/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "files": [],
+  "references": [
+    { "path": "./tsconfig.app.json" },
+    { "path": "./tsconfig.node.json" }
```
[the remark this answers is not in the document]
demariadaniel: This is Vite React boilerplate. It automatically sets up the React project this way. I can look into this tomorrow but assumed it was a pre-req to get React up and running. I don't see how the tsconfig.node is used other than for the Vite config. (see `tsconfig.node` -> `include`)

apps/ui/tsconfig.node.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "lib": ["ES2023"],
```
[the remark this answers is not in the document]
demariadaniel: ES2022 / ES2023 here are Vite template boilerplate. I can't find documentation giving an explicit reason for these vs. ESNext. I assume it's to use a modern but stable version, without introducing experimental features.

apps/ui/tsconfig.json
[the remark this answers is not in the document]
demariadaniel: Combined into a single file, and the `run dev` script still works.

packages/types/vite.config.ts
demariadaniel: Eh there was some back and forth during setup. This package, I think I chose the wrong Vite template to start with and have since removed these options.

README.md
```diff
@@ -1 +1,67 @@
...
+- [Node.js](https://nodejs.org/en) Runtime environment (v20 or higher)
+- [VS Code](https://code.visualstudio.com/) As recommended code editor. Plugins recommended: ESLint, Prettier - Code formatter, Mocha Test Explorer, Monorepo Workspace
+
+### System Dependencies
+
+### Setup
+
+## Documentation
+
+- [Postgres Database](https://www.postgresql.org/) For data storage
```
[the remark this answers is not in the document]
demariadaniel: Updated dependencies & setup

apps/ui/README.md
demariadaniel: Removed auto-generated Vite content until this is resolved

apps/api/src/main.mts
```diff
@@ -0,0 +1,42 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import cors from 'cors';
+import express, { Request, Response } from 'express';
+
+export const port = process.env.PORT;
+
+// TODO: Fix Types package so we can import from main instead of specific file
+import { demoApplication } from 'pcgl-daco/packages/data-model/src/main.mts';
```
ciaranschutte, line 26: keep imports at top of file

apps/api/src/main.mts
```diff
@@ -0,0 +1,42 @@
...
+
+const app = express();
+
+app.use(cors());
+
+app.get('/', (_req: Request, res: Response) => {
+	res.send('Hello World!');
+});
+
+app.get('/applications', cors(), (_req: Request, res: Response) => {
```
ciaranschutte, line 36: setting `cors` middleware at app level means it's enabled for all app handlers
https://expressjs.com/en/resources/middleware/cors.html
I think the first and second example both made it in

apps/api/src/tests/index.mjs
```diff
@@ -0,0 +1,35 @@
...
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import assert from 'node:assert';
+import { after, describe, it } from 'node:test';
+
+import { port, server } from '../../dist/pcgl-daco-api.js';
+
+describe('Initial Test Setup', async () => {
```
ciaranschutte, line 25: no async operation here?

apps/api/vite.config.ts
```diff
@@ -0,0 +1,23 @@
...
+			external: ['express'],
+			output: {
+				globals: {
+					express: 'express',
+				},
+			},
+		},
+	},
+	plugins: [],
+});
```
ciaranschutte, line 23: vite config for lib. doesn't make sense to me here

apps/ui/src/App.tsx
```diff
@@ -0,0 +1,61 @@
...
+		setApplication(application);
+	};
+
+	const applicant = application?.applicant;
+
+	const applicantName = applicant
+		? `${applicant.title}
+	${applicant.first} ${applicant.middle}
+	${applicant.last}`
+		: '';
```
ciaranschutte, line 40: this makes me wonder if prettier + husky is working correctly.

apps/ui/src/App.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { type Application } from 'pcgl-daco/packages/data-model/src/types';
+import { useState } from 'react';
+import './App.css';
+
+function App() {
```
ciaranschutte, line 24: good to have consistent code style throughout a project! 
We usually use `const App = () => {}` style but I'm open to this stylistic choice, let's keep it consistent for the components to come too

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+		"lib": ["ESNext"],
+		"skipLibCheck": true,
+		"allowImportingTsExtensions": true,
+		"allowSyntheticDefaultImports": true,
+		"isolatedModules": true,
+		"moduleDetection": "force",
+		"noEmit": true,
+		"baseUrl": ".",
+		"paths": {
+			"*": ["./node_modules/*"],
```
ciaranschutte, line 13: won't this remap all imports to a subdir of node_modules?

apps/api/src/main.mts
ciaranschutte: why export anything? It's server entry point

ciaranschutte:
A few comments/bits of advice. Overall looking good :thumbs-up:
I'll leave it up to you & others on the project to decide how actionable these are.

- env <br/>
https://github.com/Pan-Canadian-Genome-Library/daco/pull/7/commits/9a14e7c1da573a4019df438feef510532ea0ff2c
.env support added here, but no .env.schema added which is helpful to give developers are starting point.


- no need to export `server`, I found that confusing. `app.listen` is the magic here.

- it's easier to give prettier a specific folder like `src` to modify rather than have a tonne of ignore options. When people add other files, they may have to add to prettier ignore. aim is sensible default behaviour. <br/><br/> I don't understand the `vite.config.js` usage for server.  I don't see it in the Vite boilerplates. I think you may be basing it off the library config I shared with you for iobio. In `iobio` the end goal was a package, a library to install and use in a project. <br/><br/>It's different here.
We don't install the server or the ui as packages, so we don't need to build them as libraries.
The simplest server setup is to install dependencies then run the server (vite). 
I believe no additional config needed... just run vite.


- `app.use(cors())` is setting up middleware cors for all requests on app level, no need to use it later on individual handlers if you set it up like this.

apps/api/src/main.mts
demariadaniel: Moved

apps/api/src/main.mts
demariadaniel: Ooh sorry I see, yeah it looks like I mixed their examples. Updated.

apps/api/src/tests/index.mjs
demariadaniel: Leftover from earlier test definitions, removed

apps/ui/src/App.tsx
demariadaniel: Which aspect? 
Running `pnpm run prettier` shows this file as `unchanged`, editor makes no additional changes on save (but will change import order, for example)

demariadaniel:
> * it's easier to give prettier a specific folder like `src` to modify rather than have a tonne of ignore options. When people add other files, they may have to add to prettier ignore. aim is sensible default behaviour. 

Good call, updated npm script to target `./packages` & `./apps`

apps/api/vite.config.ts
demariadaniel: Yeah I had a feeling this wasn't necessary, let me remove

demariadaniel:
> I don't understand the `vite.config.js` usage for server.  I don't see it in the Vite boilerplates. I think you may be basing it off the library config I shared with you for iobio. In `iobio` the end goal was a package, a library to install and use in a project. It's different here.
>   We don't install the server or the ui as packages, so we don't need to build them as libraries.
>   The simplest server setup is to install dependencies then run the server (vite).
>   I believe no additional config needed... just run vite.

So given we don't need to build a package, I don't know if we need Vite at all for the API. 
I used Vite for setup but it's actually just a Node/Express app.
We could make a `run:prod` script which essentially does the same thing as `run:dev` at this stage.
This is essentially how platform-api is built: https://github.com/icgc-argo/platform-api/blob/develop/package.json#L15

> * right now the dev script is using tsx with an env flag. for prod, looks like vite should be used, so you'll likely have to use something like https://www.npmjs.com/package/dotenv to read env vars.

The flag is a native Node flag: https://nodejs.org/en/learn/command-line/how-to-read-environment-variables-from-nodejs 
I wanted to leverage modern Node features as much as possible to keep it minimal. See above comment, if Vite is not a requirement then is DotEnv needed?

tsconfig.json
demariadaniel: Good catch, previously had added this to the API config when I was still getting things running. Isn't needed now. Removed.

apps/api/src/main.mts
demariadaniel: Triggering the test runner starts the server, so I needed a reference in the test file to close it.
I've removed the export and replaced it with `process.exit()` in the tests.
We can add server-endpoint testing when needed.

apps/ui/src/App.tsx
ciaranschutte: I missed that it's a template string with newlines.
just a really odd variable I guess.

ciaranschutte (approved):
Looks good. Approving.

Not sure why there's a `docs` folder with empty files. 
It's a nice thought but unless it's definitely going to worked it's just adding empty files to a repo.
if keeping , might be worth adding some content even if very minimal for now.

demariadaniel:
> Looks good. Approving.
> 
> Not sure why there's a `docs` folder with empty files. It's a nice thought but unless it's definitely going to worked it's just adding empty files to a repo. if keeping , might be worth adding some content even if very minimal for now.

Thanks @ciaranschutte there's a handful of Readme items to resolve so I created a follow up ticket: https://app.zenhub.com/workspaces/pcgl---pancanadian-genome-library-659eed8bc4fe0c04890cff77/issues/gh/pan-canadian-genome-library/daco/14

These are template files from Overture, will revisit

Issue #14, opened 2024-10-23
Follow Up Docs Updates


demariadaniel:
A) Revise project summary based off existing project resources
https://oicr.slack.com/archives/C066W5703RU/p1729525647009309

B) Review OHCRN docs for Readme updates:
https://github.com/Pan-Canadian-Genome-Library/daco/pull/7#discussion_r1797349729

C) Remove (or update) Overture template docs accordingly:
https://github.com/Pan-Canadian-Genome-Library/daco/pull/7#pullrequestreview-2386149183

Buwujiu:
@demariadaniel You can use the following as project summary: Canada boasts world-leading expertise in genomics, including developing data-sharing policies and tools. However, we lack a national strategy to aggregate, store and share Canadian data equitably, securely and sustainably. At the same time, the size and complexity of human genomics datasets and their associated clinical data are growing rapidly. The Pan-Canadian Genome Library (PCGL) is a large collaborative effort to unify Canada's genome sequencing efforts. The PCGL is an open-source and open-science initiative, building upon Canadian-made foundational components and datasets, and utilizing international standards such as GA4GH to unify Canada’s human genome sequencing efforts.

Buwujiu:
PR merged, closing.

PR #20, opened 2024-10-29
📜 #14 Docs Updates

demariadaniel:
Issue: #14 

- Adds project summary
- Adds detail to set up sections & default .env values
- Removes Overture template files
- Some .md formatting


README.md
```diff
@@ -39,13 +45,27 @@ The modules in the monorepo are organized into two categories:
 - Install dependencies: `pnpm i`
 
 - Start server:
+  
   `cd apps/api`
+  
+  `touch .env`
+
+  Add `.env` values
+
```
ciaranschutte, line 54: consider using `cp .env.schema .env` to make a copy of the default env schema.
User then has the default schema as a base and can adjust accordingly.

demariadaniel: Amazing, had a feeling this could be improved

README.md
```diff
@@ -1,6 +1,12 @@
 # Pan Canadian Genome Library Data Access Compliance Office
 
-> <img src="./apps/ui/public/pcgl-logo.png" height="90" />
+<div>
```
justincorrigible, line 3: sheer curiosity: how does this div make any difference?

apps/api/README.md
```diff
@@ -1,7 +1,29 @@
 # PCGL DACO API
 
-Server Application for the Pan Canadian Genome Library Data Access Compliance Office
+Nodes.js API for the Pan Canadian Genome Library Data Access Compliance Office.
+
+This app uses features in Node ^20.9, in combination with TSX for TypeScript support.
+
+It provides an Express API for accessing a Postgres database using Drizzle ORM.
```
justincorrigible, line 7:
```suggestion
This app uses features from Node ^20.9, in combination with TSX for TypeScript support, and it provides access to a Postgres database using Drizzle ORM.
```

apps/api/drizzle.config.ts
```diff
@@ -19,13 +19,13 @@
...
 import { defineConfig } from 'drizzle-kit';
 
-const PGDATABASE = process.env.PGDATABASE;
-const PGUSER = process.env.PGUSER;
-const PGPASSWORD = process.env.PGPASSWORD;
-const PGHOST = process.env.PGHOST;
+const PG_DATABASE = process.env.PG_DATABASE || 'postgres';
+const PG_USER = process.env.PG_USER || 'postgres';
+const PG_PASSWORD = process.env.PG_PASSWORD || 'password';
+const PG_HOST = process.env.PG_HOST || 'localhost:5432';
```
justincorrigible, line 25: inclined to suggest we don't offer defaults/fallback (i.e. leave them as empty strings), and error out instead if these are missing (i.e. if string.length < 1)

apps/ui/.env.schema
```diff
@@ -1 +1 @@
-VITE_APPLICATION_API_URL=
\ No newline at end of file
+VITE_APPLICATION_API_URL=http://localhost:3000
```
justincorrigible, line 1: always useful to have an empty line at the end of every file

```suggestion
VITE_APPLICATION_API_URL=http://localhost:3000

```

README.md
demariadaniel: It's implemented this way in Stage: https://github.com/overture-stack/stage/blob/develop/README.md?plain=1

TBH I don't believe it makes a layout difference? At least not in VSCode Markdown preview

apps/api/drizzle.config.ts
demariadaniel: Yes, that is a great point

demariadaniel: OK great, removed

README.md
justincorrigible: how curious! if it doesn't add, let's remove it then

demariadaniel: Removed

README.md
```diff
@@ -39,13 +43,23 @@ The modules in the monorepo are organized into two categories:
...
 
 - Start UI:
+
   `cd apps/ui`
+
+  `cp .env.schema .env`
+
   `pnpm run dev`
+
   Visit `http://localhost:5173/` in the browser to get started
```
justincorrigible, line 63: is that port chosen at random? seems atypical enough to be purposefully chosen without it following a known standard

apps/api/README.md
```diff
@@ -1,7 +1,27 @@
...
+| `PG_HOST`                                | URL for the Postgres DB service. application                                                                                             | `string` | Optional | `localhost:5432`                                                                                                                                        |
+| `PG_DATABASE`                              | Postgres Database name. application                                                                                           | `string` | Optional | `postgres`                                                                                                                                        |
+
+### Unit Testing
+
+- Unit tests are built using the [Node Test Runner](https://nodejs.org/api/test.html) and are triggered using `pnpm run test`
+
+### Database
+
+- This application implements a Postgres database using the following [data model](../../docs/model/README.md)
```
justincorrigible, line 27:
```suggestion
- This application implements a Postgres database using the following [data model](../../docs/model/README.md)

```

apps/api/drizzle.config.ts
```diff
@@ -19,13 +19,13 @@
...
-const PGPASSWORD = process.env.PGPASSWORD;
-const PGHOST = process.env.PGHOST;
+const PG_DATABASE = process.env.PG_DATABASE;
+const PG_USER = process.env.PG_USER;
+const PG_PASSWORD = process.env.PG_PASSWORD;
+const PG_HOST = process.env.PG_HOST;
 
 // PG Connection String: postgres://postgres:mypassword@localhost:5432/postgres
-const connectionString = `postgres://${PGUSER}:${PGPASSWORD}@${PGHOST}/${PGDATABASE}`;
+const connectionString = `postgres://${PG_USER}:${PG_PASSWORD}@${PG_HOST}/${PG_DATABASE}`;
```
justincorrigible, line 28: not for this PR, but we could add some error handling here, so that we don't even bother making a request to the db if any of these details are missing

justincorrigible (approved):

README.md
demariadaniel: This is a Vite default: https://vite.dev/config/server-options#server-port
So this may end up being configured differently (and readme updated accordingly) or we could add a link to the docs

apps/api/drizzle.config.ts
demariadaniel: Good call, I could add this to the db setup branch I'm working on

Issue #6, opened 2024-09-26
Create Application Service Layer - Initial Simple Actions


Buwujiu:
- Create an applicationService file with the following exported functions
   - [ ] create new application
      - starts in draft state
      - creates empty application-content record
      - requires user ID as input
   -  [ ] get application
      - gets application by ID
      - includes application-content
   - [ ] list applications
      - only includes application table details (state, user, created at)
      - may be expanded in future to include summary of number of collaborators and application content progress, don't need this now
      - [ ] filters:
         - [ ] by user ID
         - [ ] by state (ie. only show daco review required)
      - [ ] can be sorted:
         - [ ] by created date, asc or desc
         - [ ] by updated date, asc or desc
         - [ ] by state - use the order the states are listed in the datamode as their ascending order (draft=1, institution-rep-review=2, etc.)
- Create test suite for application service
   - [ ] Integration tests that use testcontainers to interact with postgres
   - [ ] create file with seed data for postgres to test read actions
- This ticket tracks service layer code only, not web routes to access this information

PR #40, opened 2024-11-12
🐕‍🦺 Feat/6 API Service Layer

demariadaniel:
https://github.com/Pan-Canadian-Genome-Library/daco/issues/6

- Adds Application Service w/ create, find by id, list many, update, delete functionality
  - this template can be adapted to other Models
- Includes filtering / sorting as described in ticket
- Full set of unit tests
- Separates implementations for connecting to live DB / test-DB
- Add `updatedAt` field to Applications
- Fix a few relation fields


apps/api/.env.test
```diff
@@ -2,7 +2,7 @@
 PORT=3000
 
 # DB
-PG_USER=postgres
-PG_PASSWORD=mypassword
-PG_HOST=localhost:5432
+PG_USER=testUser
+PG_PASSWORD=testPassword
 PG_DATABASE=postgres
+PG_HOST=localhost:5000
```
demariadaniel, line 8: Just differentiating from .env.schema values

apps/api/src/service/application-service.ts
```diff
@@ -0,0 +1,164 @@
...
+			console.log(`Application created with user_id: ${user_id}`);
+
+			return application[0];
+		} catch (err) {
+			console.error(`Error at createApplication with user_id: ${user_id}`);
+			console.error(err);
+			return null;
+		}
+	},
+	findOneAndUpdate: async ({ id, update }: { id: number; update: any }) => {
```
demariadaniel, line 60: Not listed in the ticket but created out of the need to update records in the tests

apps/api/src/service/application-service.ts
```diff
@@ -0,0 +1,164 @@
...
+				.orderBy(sortQuery);
+
+			return allApplications;
+		} catch (err) {
+			console.error(`Error at listApplications with user_id: ${user_id} state: ${state}`);
+			console.error(err);
+			return null;
+		}
+	},
+	deleteApplication: async ({ user_id }: { user_id: string }) => {
```
demariadaniel, line 150: Similar to `findOneAndUpdate`, there was a need to delete records after they were created.

There were connection issues (now fixed) creating two separate DB connections (one for the service, one for the tests) so I needed additional service methods first to resolve the test issues before I diagnosed the connection issues.

apps/api/src/service/application-service.ts
```diff
@@ -0,0 +1,164 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { and, asc, desc, eq, sql } from 'drizzle-orm';
+import { ApplicationStates } from 'pcgl-daco/packages/data-model/src/types.ts';
+import { type PostgresDb } from '../db/index.ts';
+import { applicationContents } from '../db/schemas/applicationContents.ts';
+import { applications } from '../db/schemas/applications.ts';
+
+const applicationService = (db: PostgresDb) => ({
```
demariadaniel, line 26: Open to alternative approaches to passing the database connection reference as an argument

apps/api/src/tests/services/application.ts
```diff
@@ -0,0 +1,164 @@
...
+
+	it('should allow sorting records by created_at', async () => {
+		const applicationRecords = await applicationService.listApplications({ sort: 'created_at' });
+
+		assert.ok(Array.isArray(applicationRecords));
+		assert.strictEqual(applicationRecords.length, 3);
+
+		const date1 = applicationRecords[0].createdAt.valueOf();
+		const date2 = applicationRecords[1].createdAt.valueOf();
+		const date3 = applicationRecords[2].createdAt.valueOf();
```
demariadaniel, line 105: Better variable names?

apps/api/src/main.ts
```diff
@@ -17,23 +17,7 @@
...
-});
-
-app.get('/applications', cors(), (_req: Request, res: Response) => {
-	res.send(demoApplication);
-});
-
-app.listen(port, () => {
-	console.log(`Example app listening on port ${port}`);
-});
+startServer(connectionString);
```
demariadaniel, line 23: Adding separation between starting server and starting test-server

apps/api/src/tests/db/index.ts
```diff
@@ -18,29 +18,35 @@
...
+
+	const user_id = 'testUser@oicr.on.ca';
+
+	before(async () => {
+		container = await new PostgreSqlContainer().start();
+		const connectionString = container.getConnectionUri();
+		db = startDb(connectionString);
+
+		await initMigration(db);
+	});
```
demariadaniel, line 47: These are tests from the previous PR which do not have a Service yet (test plain Drizzle actions)

JamesTLopez (commented):
Just some debugging console.logs and perhaps naming improvments with the docker file

apps/api/src/service/application-service.ts
```diff
@@ -42,21 +42,15 @@ export const applicationService = {
...
+				.set({ contents: contentsId })
 				.where(eq(applications.id, id))
-				.leftJoin(applicationContents, eq(applications.contents, applicationContents.id));
+				.returning();
 
-			const application = {
-				...applicationRecord[0].applications,
-				contents: applicationRecord[0].application_contents,
-			};
+			console.log(`Application created with user_id: ${user_id}`);
```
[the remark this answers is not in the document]
demariadaniel: Yeah fair point, this loop is definitely just a placeholder for a batch insert, or populating the test DB on init. Could be improved for sure, added that pretty quick at the end of the day yesterday.

However, the service level logs were intentional; there is a decent amount of logging in our Argo Clinical Service indicating which endpoints were called with what arguments.
So I think a base level of reporting is beneficial too.

examples:
Screenshot 2024-11-15 at 2 29 36 PM

Screenshot 2024-11-15 at 2 24 59 PM

apps/api/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+services:
+  postgres:
```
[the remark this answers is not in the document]
demariadaniel: Good call, added

apps/api/src/service/application-service.ts
```diff
@@ -59,8 +59,15 @@ export const applicationService = {
 		return allApplications;
 	},
 	deleteApplication: async ({ user_id }: { user_id: string }) => {
-		const deletedRecords = await db.delete(applications).where(eq(applications.user_id, user_id)).returning();
+		try {
+			const deletedRecords = await db.delete(applications).where(eq(applications.user_id, user_id)).returning();
+			console.log(`Application deleted with user_id: ${user_id}`);
```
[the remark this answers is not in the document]
demariadaniel: Commented on: https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=87273850&issue=Pan-Canadian-Genome-Library%7Cdaco%7C44

We will standardize logging when we get to the REST side of things

apps/api/src/service/application-service.ts
```diff
@@ -0,0 +1,167 @@
...
+		pageSize = 20,
+	}: {
+		user_id?: string;
+		state?: ApplicationStates;
+		sort?: string;
+		page?: number;
+		pageSize?: number;
+	}) => {
+		const isDescending = sort?.charAt(0) === '-';
+		const sortValue = isDescending ? sort.substring(1) : sort;
```
joneubank, line 107: This encodes a lot of rules for the content of the "sort" string value, and our type system won't validate if they are formatted correctly or contain valid fields for our data model. If we do use a string like this, we should add TSDoc comments on this function to ensure the content of this parameter are well documented. I have an alternate suggestion though:

I see the `-columnName` pattern for sort parameters used when encoded in URL query params, where you want to put condensed information and only string values. We don't have these restriction on our internal functions parameters. We can explicitly require our sort paramater to have two properties, one for direction and the other for the column name, and the values for these properties can be strictly set as `asc | desc` and as the list of legal column names for the schema you are sorting.

It would look something like:

```ts
type OrderBy<T extends string> = {
	direction: 'asc' | 'desc';
	column: T;
};
type ApplicationsColumnName = keyof typeof applications.$inferSelect;

// the sort function argument can be typed as:
(sort?: Array<OrderBy<ApplicationsColumnName>>) => { 

// Making this an array allows the function to orderBy multiple properties, sorting by user ascending and then created_at descending, for instance.

// The orderBy clause can be built like:
const orderByArguments = sort
	? sort.map((sortBy) =>
			sortBy.direction === 'asc' ? asc(applications[sortBy.column]) : desc(applications[sortBy.column]),
	  )
	: [asc(applications.created_at)];

	await db.select({}).orderBy(...orderByArguments);
}
```

Lastly, if we expect to do db sorting in multiple places, then we could build some of this code as common utilities, for sort types and to built the orderby arguments.

apps/api/src/service/application-service.ts
joneubank: We have no use case for deleting applications. In order to prevent this being used when we shouldnt, can we remove this function?

apps/api/src/tests/services/application.ts
```diff
@@ -0,0 +1,197 @@
...
+		const { id } = applicationRecords[0];
+
+		const requestedApplication = await applicationService.getApplicationById({ id });
+
+		assert.notEqual(requestedApplication, null);
+		assert.strictEqual(requestedApplication?.id, id);
+		assert.strictEqual(requestedApplication?.id, requestedApplication?.contents?.application_id);
+	});
+
+	it('should list applications filtered by user_id', async () => {
```
joneubank, line 81: The next set of tests are for the listApplications, can we group them into a `suite('listApplications', async ()=>{...})`?

apps/api/src/tests/services/application.ts
```diff
@@ -0,0 +1,197 @@
...
+		assert.ok(Array.isArray(applicationRecords));
+		assert.strictEqual(applicationRecords.length, 3);
+	});
+
+	it('should list applications filtered by state', async () => {
+		const applicationRecords = await applicationService.listApplications({ state: ApplicationStates.DRAFT });
+
+		assert.ok(Array.isArray(applicationRecords));
+		assert.strictEqual(applicationRecords.length, 3);
+	});
```
joneubank, line 96: How do you initialize the data for this test? It looks like it depends on previous testa and that if I run this test directly it will fail. We want our unit tests to be repeatable and standalone, able to run in any order or even on their own.

If there is a series of steps that we can't separate to test independently, and would be valuable to test in sequence, then we should build these as a single test with sub tests. That way we are assured to run them in sequence.

joneubank: This comment applies to many tests in this file. We should have an script that inserts a test data set before running the tests that read data. We can use our create application api to insert these records.

apps/api/src/tests/services/application.ts
```diff
@@ -0,0 +1,197 @@
...
+
+	it('should list applications filtered by state', async () => {
+		const applicationRecords = await applicationService.listApplications({ state: ApplicationStates.DRAFT });
+
+		assert.ok(Array.isArray(applicationRecords));
+		assert.strictEqual(applicationRecords.length, 3);
```
joneubank, line 95: It may make more sense to ensure that every record in the response has the correct state, instead of just checking the count... unless you have cleared the db and initialized it with a precise test data set.

apps/api/src/tests/services/application.ts
joneubank: This is fine really.

apps/api/src/tests/services/application.ts
```diff
@@ -0,0 +1,197 @@
...
+		const applicationRecords = await applicationService.listApplications({ user_id });
+
+		assert.ok(Array.isArray(applicationRecords));
+		assert.strictEqual(applicationRecords.length, 3);
+
+		const { id: zeroRecordId } = applicationRecords[0];
+		const { id: firstRecordId } = applicationRecords[1];
+		const { id: secondRecordId } = applicationRecords[2];
+
+		// State values are used in the next test so first record remains in 'Draft', this is just populating `updatedAt`
```
joneubank, line 122: TODO: Tests need to be independent or grouped together as a single test with sub tests.

joneubank (changes requested):
Solid foundation here, lots of smart additions. Thanks!

apps/api/.env.test
joneubank: On second review, that example may not work with our tests, since the env might be set by the `before()` function after the env is read by the startDb function... Need to test and confirm, I have ideas if need be. I stand by the idea that we shouldn't be feeding an env file into our test suite except to configure/control what tests are running or how they are being run.

apps/api/src/tests/services/application.ts
demariadaniel: Yes definitely. I parked some of the test setup tasks because Drizzle seeding is not available yet (see comment at line 56 of this file).

The tests are written sequentially (so that Test B depends on Test A etc.) but this is not ideal (or sustainable), and I intended to update this before we went much further.

I opened the PR at this state because it was functionally complete, and I wanted to take a step back to assess how we should approach test setup with the team's input.

apps/api/src/tests/services/application.ts
demariadaniel: See below comment -- I'm aware the test setup could use improvements and felt that was one piece worth revisiting. I should have expanded on my notes so you were aware too!

apps/api/drizzle/0000_nosy_echo.sql
```diff
@@ -70,6 +70,7 @@ CREATE TABLE IF NOT EXISTS "applications" (
 	"state" "application_states" NOT NULL,
 	"created_at" timestamp DEFAULT now() NOT NULL,
 	"approved_at" timestamp,
+	"updated_at" timestamp,
 	"expires_at" timestamp,
 	"contents" bigint
```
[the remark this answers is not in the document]
demariadaniel: Yes agreed, I was thinking we'd be adding migration files as soon as there's anything deployed. Very good point about informing rest of team re: running migrations.

apps/api/drizzle/0000_nosy_echo.sql
leoraba, line 1: I think you can generate migration files and give a more meaningful name using `npx drizzle-kit generate --name=init` 
doc:  https://orm.drizzle.team/docs/drizzle-kit-generate#custom-migration-file-name

leoraba: or just rename it manually in 😜

apps/api/src/db/schemas/applications.ts
leoraba, line 1: This schema change is not reflected in the docs. May I suggest a tool to generate .dbml files e.g. [drizzle-dbml-generator](https://www.npmjs.com/package/drizzle-dbml-generator)
That task can be added in package.json on build time for example to generate .dbml and keep the docs updated

leoraba (commented):
LGTM. just left some comments 👍

apps/api/src/db/index.ts
```diff
@@ -0,0 +1,50 @@
...
+import { dirname } from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const migrationsFolder = __dirname + '/../../drizzle';
+
+export type PostgresDb = ReturnType<typeof drizzle>;
+
+export const startDb = (connectionString: string): PostgresDb => {
```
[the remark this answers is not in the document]
demariadaniel: All good, improved to use connectToDb, important distinction

apps/api/src/service/application-service.ts
demariadaniel: Removed

apps/api/src/tests/services/application.ts
demariadaniel: Yes, added a `describe()` which is just alias for suite, later we can add same for create/get etc.

apps/api/src/service/application-service.ts
```diff
@@ -0,0 +1,167 @@
...
+		try {
+			const newApplicationRecord = await db.insert(applications).values(newApplication).returning();
+			const { id } = newApplicationRecord[0];
+
+			const newAppContents: typeof applicationContents.$inferInsert = {
+				application_id: id,
+				created_at: new Date(),
+				updated_at: new Date(),
+			};
+			const newAppContentsRecord = await db.insert(applicationContents).values(newAppContents).returning();
```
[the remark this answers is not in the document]
demariadaniel: Yep, I've grouped these in a transaction

apps/api/src/service/application-service.ts
```diff
@@ -0,0 +1,167 @@
...
+			console.error(`Error at createApplication with user_id: ${user_id}`);
+			console.error(err);
+			return null;
+		}
+	},
+	findOneAndUpdate: async ({ id, update }: { id: number; update: any }) => {
+		try {
+			const application = await db
+				.update(applications)
+				.set({ ...update, updated_at: sql`NOW()` })
```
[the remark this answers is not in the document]
demariadaniel: No, so I've added one to start the FindOneAndUpdate suite

apps/api/src/service/application-service.ts
demariadaniel: Update/set types are not well documented but finally figured this out

apps/api/src/service/application-service.ts
demariadaniel: Also added types/utils files

apps/api/.env.test
demariadaniel: Setting process.env does work but it's not actually necessary because we can just set the container variables and obtain the connection string like so:

```
container = await new PostgreSqlContainer()
			.withUsername(PG_USER)
			.withPassword(PG_PASSWORD)
			.withDatabase(PG_DATABASE)
			.start();

const connectionString = container.getConnectionUri();
db = connectToDb(connectionString);
```

where PG_USER etc. are defined as constants in a test file

apps/api/src/db/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const initMigration = async (db: PostgresDb) => {
+	try {
+		await migrate(db, { migrationsFolder });
+	} catch (err) {
+		console.log('Error Migrating on Database startup');
+		console.log(err);
+		throw err;
+	}
+};
```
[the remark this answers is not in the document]
demariadaniel: Moved to testUtils

apps/api/src/tests/services/application.ts
demariadaniel: Updated to confirm state values, as well above user_id test now checks user_id

apps/api/src/tests/services/application.ts
demariadaniel: Added test-data init functions to testUtils

apps/api/src/tests/services/application.ts
demariadaniel: Tests are now grouped together and setup functions enable independent execution

apps/api/src/db/schemas/applications.ts
demariadaniel: OK I added a script to generate the DBML, using schema generated by `drizzle-kit pull`
The drizzle-config is using a glob path string which doesn't work with pgGenerate
So rather than refactor all the schema / table / enum / relations code `pull` works

However, there is a conflict w/ `files.content` field which will need to be revisited
I've left a TODO comment and this can be reviewed when we do work on Files service

packages/data-model/src/types.ts
```diff
@@ -20,14 +20,14 @@
...
-	'CLOSED',
-	'REVOKED',
+	DRAFT = 'DRAFT',
+	INSTITUTIONAL_REP_REVIEW = 'INSTITUTIONAL_REP_REVIEW',
+	DAC_REVIEW = 'DAC_REVIEW',
+	DAC_REVISIONS_REQUESTED = 'DAC_REVISIONS_REQUESTED',
+	REJECTED = 'REJECTED',
+	APPROVED = 'APPROVED',
+	CLOSED = 'CLOSED',
+	REVOKED = 'REVOKED',
```
joneubank, line 30: Important to use these named enum values if we are using TS enums, so thank you for this update. We can take a moment to review our enum usage in a future change.

apps/api/drizzle/schema.ts
```diff
@@ -0,0 +1,217 @@
...
+});
+
+export const agreements = pgTable('agreements', {
+	// You can use { mode: "bigint" } if numbers are exceeding js number limitations
+	id: bigint({ mode: 'number' }).primaryKey().generatedAlwaysAsIdentity({
+		name: 'agreements_id_seq',
+		startWith: 1,
+		increment: 1,
+		minValue: 1,
+		maxValue: 9223372036854775807,
```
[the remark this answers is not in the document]
demariadaniel: This was generated using `drizzle-kit pull` so it's the auto-generated upper limit for a BigInt

apps/api/src/service/application-service.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const applicationService = (db: PostgresDb) => ({
+	createApplication: async ({ user_id }: { user_id: string }) => {
+		const newApplication: typeof applications.$inferInsert = {
+			user_id,
+			state: ApplicationStates.DRAFT,
+		};
+
+		try {
+			const application = await db.transaction(async (transaction) => {
```
leoraba, line 36: 👍

apps/api/drizzle/schema.ts
```diff
@@ -0,0 +1,217 @@
...
+	institutionalRepPrimaryAffiliation: varchar('institutional_rep_primary_affiliation', { length: 255 }),
+	institutionalRepEmail: varchar('institutional_rep_email', { length: 255 }),
+	institutionalRepProfileUrl: varchar('institutional_rep_profile_url', { length: 255 }),
+	institutionalRepPositionTitle: varchar('institutional_rep_position_title', { length: 255 }),
+	institutionCountry: varchar('institution_country', { length: 255 }),
+	institutionState: varchar('institution_state', { length: 255 }),
+	institutionCity: varchar('institution_city', { length: 255 }),
+	institutionStreetAddress: text('institution_street_address'),
+	institutionPostalCode: varchar('institution_postal_code', { length: 255 }),
+	institutionBuilding: varchar('institution_building', { length: 255 }),
```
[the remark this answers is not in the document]
demariadaniel: `If character varying (or varchar) is used without length specifier, the type accepts strings of any length.`
https://www.postgresql.org/docs/current/datatype-character.html
https://orm.drizzle.team/docs/column-types/pg#varchar
Drizzle says nothing so referring to PG docs

apps/api/src/tests/index.ts
leoraba, line 1: just wondering, should `test/` folder be at top level along the `src/`. also consider changing extension to differentiate the actual code like `.test.ts` or `.spec.ts`

demariadaniel: True, argo-clinical does have `test` at the same level as `src` and we have generally used the `.spec` convention

I need these changes in `main` to unblock other work so I can incorporate these suggestions into the next branch

leoraba (approved):
LGTM 🎉

apps/api/drizzle/schema.ts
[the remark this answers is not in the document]
demariadaniel: Leo and I chatted about this -- there is a disconnect between the DBML schema input (expects an object) and the Drizzle config schema definition (string glob path to schemas folder './schemas/**')

This should be refactored so that both point to the same schema definition, my only hesitation is to limit the scope of this PR. I will update this and the test files in the next branch.

PR #67, opened 2024-11-20
Feat/51 applicationspage empty

JamesTLopez:
- add react router
- create dashboardpage
- create layout wrappers for statusbar and content
- create new application card

### DIFF
--- a/apps/ui/src/components/ApplicationStatusBar.tsx
+++ b/apps/ui/src/components/ApplicationStatusBar.tsx
@@ -26,11 +26,11 @@ const { Text, Title } = Typography;
 const { useToken } = theme;
 
 // TODO: properly type and verify incoming data once API service is completed
-interface ApplicationStatusBarType {
-	expires_at?: string;
-}
+type ApplicationStatusBarProps = {
+	expiresAt?: string;
+};
 
-const ApplicationStatusBar = ({ expires_at }: ApplicationStatusBarType) => {
+const ApplicationStatusBar = ({ expiresAt }: ApplicationStatusBarProps) => {
 	const { token } = useToken();
 
 	return (
@@ -57,12 +57,12 @@ const ApplicationStatusBar = ({ expires_at }: ApplicationStatusBarType) => {
 							align="center"
 							gap={20}
 						>
-							{expires_at ? (
+							{expiresAt ? (
 								<>
 									<CheckCircleFilled style={{ color: token.colorPrimary, fontSize: 30 }} />
 									<Flex vertical>
 										<Text strong>You have access to PCGL Controlled Data</Text>
-										<Text>Expires: {expires_at}</Text>
+										<Text>Expires: {expiresAt}</Text>
 									</Flex>
 								</>
 							) : (
--- a/apps/ui/src/components/ApplicationStatusBar.tsx
+++ b/apps/ui/src/components/ApplicationStatusBar.tsx
@@ -18,7 +18,7 @@
  */
 
 import { CheckCircleFilled, CloseCircleFilled } from '@ant-design/icons';
-import { Col, Flex, Typography, theme } from 'antd';
+import { Col, Flex, Row, Typography, theme } from 'antd';
 
 import StatusBarWrapper from '@/components/layouts/StatusBarWrapper';
 
@@ -35,7 +35,7 @@ const ApplicationStatusBar = ({ expiresAt }: ApplicationStatusBarProps) => {
 
 	return (
 		<StatusBarWrapper>
-			<>
+			<Row style={{ width: '100%' }}>
 				<Col span={12}>
 					<Flex style={{ height: '100%' }} vertical justify="center">
 						<Title>My Applications</Title>
@@ -74,7 +74,7 @@ const ApplicationStatusBar = ({ expiresAt }: ApplicationStatusBarProps) => {
 						</Flex>
 					</Flex>
 				</Col>
-			</>
+			</Row>
 		</StatusBarWrapper>
 	);
 };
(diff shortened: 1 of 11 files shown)


apps/ui/src/components/ApplicationStatusBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Col, Flex, Typography, theme } from 'antd';
+
+import StatusBarWrapper from '@/components/layouts/StatusBarWrapper';
+
+const { Text, Title } = Typography;
+const { useToken } = theme;
+
+// TODO: properly type and verify incoming data once API service is completed
+interface ApplicationStatusBarType {
+	expireDate?: string;
```
demariadaniel, line 30: This is currently `expires_at` in the database, potentially expiresAt

apps/ui/src/components/providers/ThemeProvider.tsx
```diff
@@ -24,17 +24,30 @@ const pcglTheme: ThemeConfig = {
 	token: {
 		fontFamily:
 			"Open Sans, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'",
+
 		colorPrimary: '#C41D7F',
-		colorBgContainer: '#FFF0F6',
+		colorFillAlter: '#FFFFFF',
```
demariadaniel, line 29: #FFFFFF is just white and this is a common enough color that I don't think we need to redefine it several times for different components. Wondering if there's a way we can do something akin to `colorFillAlter: pcglTheme.color.white`

demariadaniel: Reviewed a little further, I'm more wondering about this name `colorFillAlter`; is it specific to Ant Design, or are we just adding a new name for white?

In StatusBannerWrapper we pass `backgroundColor: token.colorFillAlter`, but since we're just setting the `backgroundColor` property to the `colorFillAlter` value, we could probably substitute `token.colorFillAlter` for `token.color.white`

JamesTLopez: Big fan of this suggestion. Refactoring themes

apps/ui/src/pages/App.tsx
```diff
@@ -18,20 +18,33 @@
  */
 
 import { Flex, Layout } from 'antd';
+import { createBrowserRouter, RouterProvider } from 'react-router-dom';
```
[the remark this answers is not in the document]
JamesTLopez: I agree, where do you think we should add the 'routes' file?

apps/ui/src/components/providers/ThemeProvider.tsx
demariadaniel: OK, I do see the names have meaning in Ant so maybe it's just friction with their approach. But when we're just setting CSS properties like `backgroundColor` then adding standard color definitions can help keep code DRY and easily readable.

JamesTLopez: So we are locked into properties set by the ThemeConfig provided by Ant design (like you suggested above), I have created a new object with our style guide colors (ones specifically that have been used in the docs) and using that object we can maintain DRY. Just pushed the changes I was thinking of initially, lmk how it is.

apps/ui/src/components/providers/ThemeProvider.tsx
```diff
@@ -19,22 +19,55 @@
...
+
+	warningPrimary: '#FAAD14',
+	warningSecondary: '#FFFFB8',
+
+	successPrimary: '#52C41A',
+	successSecondary: '#D9F7BE',
+
+	white: '#FFFFFF',
+	black: '#000000',
+};
```
demariadaniel, line 40: Yes this looks great 👍

apps/ui/src/pages/App.tsx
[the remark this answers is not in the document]
demariadaniel: OK giving this some thought. I'm comparing with our Next.js projects but that model doesn't necessarily apply 100% here.

Maybe our first move is to export route definitions from the `/pages` files, and move the React UI into `/components/pages`.

Comparing with our Next.js apps, the structure is essentially:

```
- components
   - providers
   - pages
      pure UI components
- pages
   - [route name]
      route component
   App
```
The files in `/components/pages/` are purely rendering / presentation layer.
The route components are concerned with Auth + Page props, and import the UI component (`route.element`).

 So I think it's fair to move Route definitions into the individual `route component` files in the `pages` folder.
 So `/pages/dashboard/index` yields a route file, essentially setting up our 'file routing' structure.
The one exception is the homepage which remains at `/pages/index` because it represents the page at `/`


Now, by this logic it might make sense to colocate our `routes` file in `/pages`.
Which just imports all `pages/[route]/index` files and exports the BrowserRouter/RouterProvider.

We do have `/components/providers`. But this is not a provider that we are likely to use for another purpose, and we're not doing anything to extend the React Router as it's own component.

So I think a) separate route logic and rendering logic and b) separate App logic from routing logic, probably at `/pages` level unless you think `/components/providers` makes more sense.

Open to ideas!

joneubank (changes requested):

apps/ui/src/components/ApplicationStatusBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface ApplicationStatusBarType {
+	expires_at?: string;
+}
+
+const ApplicationStatusBar = ({ expires_at }: ApplicationStatusBarType) => {
+	const { token } = useToken();
+
+	return (
+		<StatusBarWrapper>
+			<>
```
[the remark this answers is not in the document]
JamesTLopez: Its supposed to be a <Row> Element from antd, this is a mistake on my part. Updated

joneubank (approved):

Issue #39, opened 2024-11-12
Set Up DEV and QA Env


Buwujiu:
## Descriptions
Set up a dev and a QA environment for DACO deploy. 
Updates - Needs to work on Jenkins set up. Jon will ask Yelizar, Will have env before holiday.

Buwujiu:
Dev is set up. Since we don't have enough features to split a stable release, we will set up QA in a few more sprints. Closing.

PR #76, opened 2024-11-26
chore: Add Dockerfile building UI and API images

joneubank:
## Summary

Adds Dockerfile that creates an image for the DACO UI and API. 

The DACO UI is build as a static site hosted on a container running a simple nginx config

The API is built and run directly by with node runtime.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/39

### daco-api
- Fix imports in files
- Fix import of workspace dependencies

### daco-ui
- Add nginx config

### data-model
- Refactor TS files to all have the extension `.ts`
- Remove vite package and config
- Add `build` script

### Monorepo
- Add Dockerfile
   - Includes a single build stage that is reused by all published images
   - Adds docker image for daco-ui which runs nginx hosting the static build of the daco-ui
   - Image docker image for daco-api which runs the compiled api app in Node 20
   - Includes .dockerignore to prevent local files from 
- Modify the base tsconfig
   - All packages emit the built JS code by default, only static sites need to prevent this emit
   - All packages produce source maps and declaration files
   - Create a separate tsconfig for apps at `./apps` which disables source maps and type declarations since those are not necessary in a running application
- Add `rimraf` as a dependency, used in all build scripts to clean up stale output directories.
- Add global npm script `build:all` that will run the build stage of every project in the monorepo.

### Future Work
This shouldn't be the final form for this setup. The daco-api image includes the entire source code and all dependencies, including dev dependencies. This is not a large issue since this image is only run on the server but it does include more files and content than strictly necessary.

Long term, the dependencies need to be pruned and restricted exclusively to the code that is running on the server. We should consider monorepo managers such as Turborepo or Nx to assist with this:

Consider: https://fintlabs.medium.com/optimized-multi-stage-docker-builds-with-turborepo-and-pnpm-for-nodejs-microservices-in-a-monorepo-c686fdcf051f

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


tsconfig.json
```diff
@@ -1,21 +1,3 @@
...
-		"strict": true,
-		"useDefineForClassFields": true
-	},
-
-	/* Linting */
-	"noUnusedLocals": true,
-	"noImplicitAny": true,
-	"noUncheckedIndexedAccess": true,
-	"noFallthroughCasesInSwitch": true
+	"extends": "../tsconfig.base.json"
```
leoraba, line 2: I guess path should point to local dir `./tsconfig.base.json`

Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+ENV APP_UID=9999
+ENV APP_GID=9999
+RUN addgroup -S -g $APP_GID ${APP_USER} \
+	&& adduser -S -u $APP_UID -g $APP_GID ${APP_USER}
+
+RUN mkdir -p ${WORKDIR} \
+	&& chown -R ${APP_USER}:${APP_USER} ${WORKDIR}
+
+# Enables pnpm
+RUN corepack enable
```
leoraba, line 34: 👍

Dockerfile
```diff
@@ -0,0 +1,103 @@
+# Global variables
+ARG COMMIT=""
+ARG APP_USER=pcgldaco
+ARG WORKDIR=/app
+
+#######################################################
+# Configure Base Image
+#######################################################
+FROM node:20-alpine AS base
```
leoraba, line 9: also can work with `node:22-alpine` as it's the current LTS.

tsconfig.json
joneubank: oops, yes, thanks.

JamesTLopez:
Looks pretty clean and readable. Maybe we should think about import alias on the api side but that would be out of scope. 
LGTM.

Issue #52, opened 2024-11-14
My Applications - Grid View


Buwujiu:
## Descriptions
Build the My Application page to list all applications per user. 
Mockup: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2090-13079&t=YRaEzo6DwZgXdCEP-4
and 
https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2667-11921&t=LQ31wzxBSA3Ao4Bm-4
![Image](https://github.com/user-attachments/assets/df94314c-6a47-443d-b1d4-dce51e40f029)

## Acceptance Criteria
Refer to https://docs.google.com/document/d/1NK5j-ZgwPDwkUmMvC3KfNIdXAhm8JrWmSWBekOtN_SM/edit?tab=t.0#heading=h.w8kjofakjd9a

## API Dependency
**For listing all applications, Might need to use dummy app data for now before APIs are ready.**

patrickdsantos:
Wonderful work team! No feedback.

Buwujiu:
Thanks @patrickdsantos for testing! Closing ticket.

PR #77, opened 2024-11-26
#52: Application Page With Content UI

JamesTLopez:
## Summary

REFER TO TICKET: [DACO-52](https://github.com/Pan-Canadian-Genome-Library/daco/issues/52)

Application dashboard page UI implementation including states of project cards using custom hook and implements routing solution using react-router v7 

## Description of Changes
- UI implementation of the dashboard page 
- routing implementation (react-router v7)
- create customHook to manage application card states
- create mock data from data-model
- create global folder with utils file (date formatter)

**Since theres is a new package please run the following:**
```
- pnpm i

```
**NOTE: There is no UI logic/link to navigate to the dashboard so please enter "/dashbaord" in the URL to see the changes on the browser**

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**


apps/ui/src/hooks/useApplicationStatusHook.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * @description Custom hook manage state of the application cards
+ *
+ * @param applicationStatus
+ * @returns showEdit - show edit button logic
+ * @returns showActionRequired - show action required text
+ * @returns color - the color to apply to to the application status
+ *
+ */
+
+const useApplicationStatusHook = (applicationStatus: ApplicationStatus) => {
```
joneubank, line 40: Do you intend to have this hook be updatable, so that if the `ApplicationStatus` changes then the state will update to the new `showEdit`, `showActionRequired`, and `color` values? This was written as a hook and uses an effect that will update if applciationStatus changes, but we don't actually store the `applicationStatus` in state and we don't expose a `setApplicationStatus(status)` function.

It seems like this is a simple function that takes in an `ApplicationStatus` and returns an object with relevant stateful properties, which would be much simpler to maintain than a custom hook that is used by a single component.

```ts
type ApplicationCardProps = {
  showEdit: boolean;
  showActionRequired: boolean;
  color: string;
}
function getApplicationStatusProperties(applicationStatus: ApplicationStatus): ApplicationStateProps {
  switch(applicationStatus) {
    // ...
  }
}
```

joneubank (changes requested):
Great work here. Comments on design of hook.

apps/ui/src/global/utils.ts
joneubank, line 1: Let's start organizing our helpful functions at this point, otherwise a generic file called "utils" will become a dumping ground.

This function is limited in scope to the ApplicationCard, so I would suggest defining it and its return type in a single file `getApplicationStatusProperties.ts`, and put that file in the same dir as the `ApplicationCard.tsx`.

joneubank (changes requested):
Think we could reorganize the files a little bit. Are the `ApplicationStatusBar`, `ApplicationCard`, `NewApplicationCard` 
used in multiple pages, or should we group the MyApplication page components together under the path `components/pages/myApplication`?

apps/ui/src/global/utils.ts
JamesTLopez: @joneubank Do you think we benefit from a pages folder inside "components" and have its components/utils live inside its respective page? For example:

```
components
│
└───pages
│   │   homepage
|   |      └ etc..
│   │   Login
|   |      └ etc..
│   │
│   └───dashboard
│       │   cards
|       |     └── 
│       │        | ApplicationCard.tsx
│       │        | NewApplicationCard.tsx
│       │   ApplicationStatusBar.tsx
│       │   getApplicationStatusProperties.ts
│       │  
```
or something similar of the sort?

joneubank (approved):

PR #83, opened 2024-12-02
#71: Swagger UI Setup With Health Check

JamesTLopez:
## Summary

Swagger UI setup and create health endpoint check

## Description of Changes

- Add swagger-ui-express package 
- add yamljs package
- create swagger file
- create health point logic
- add swagger endpoint to server
- add health point check for db 

### Special Instructions
Before running these changes, you will need to install `swagger-ui-express` , `yamljs`, `@types/yamljs` and `@types/swagger-ui-express`:

```
pnpm i
```

## Readiness Checklist

- [ ] **Self Review**
- [ ] **PR Format**
- [ ] **Labels Added**
- [ ] **Local Testing**
- [ ] Documentation

### DIFF
--- a/apps/api/src/app-health.ts
+++ b/apps/api/src/app-health.ts
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2024 The Ontario Institute for Cancer Research. All rights reserved
+ *
+ * This program and the accompanying materials are made available under the terms of
+ * the GNU Affero General Public License v3.0. You should have received a copy of the
+ * GNU Affero General Public License along with this program.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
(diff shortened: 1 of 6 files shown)


apps/api/src/app-health.ts
```diff
@@ -0,0 +1,82 @@
...
+export enum Status {
+	OK = '😇',
+	UNKNOWN = '🤔',
+	ERROR = '😱',
+}
+
+export type ComponentStatus = {
+	status: Status;
+	statusText?: 'OK' | 'N/A' | 'ERROR';
+	info?: any;
```
demariadaniel, line 35: This is following the example set in argo-clinical but I think we could improve on `info?: any` going forward

The current usage is `info: { err: unknown}`, we could update this type to reflect that, then expand from there

apps/api/src/server.ts
```diff
@@ -20,16 +20,42 @@
...
+		// TODO: implement proper versioning logic when in place
+		const resBody = {
+			version: `PCGL-V1`,
+			health,
+		};
+		if (health.all.status == Status.OK) {
+			res.send(resBody);
+			return;
+		}
+		res.send(resBody);
```
demariadaniel, line 56: What is the benefit of the `if` statement? In both cases `res.send(resBody)` gets called then the function exits

JamesTLopez: Woops nice catch, I was debugging the response object and removed the last functionality and forgot to add it back in: `res.status(500).send(resBody)` and `res.status(200).send(resBody)` is what it was before.

apps/api/src/db/index.ts
```diff
@@ -18,17 +18,20 @@
...
 	try {
 		const db = drizzle(connectionString);
 
+		setStatus('db', { status: Status.OK });
 		return db;
 	} catch (err) {
 		console.log('Error on Database startup');
 		console.log(err);
+		setStatus('db', { status: Status.ERROR, info: { err } });
 		throw err;
```
demariadaniel, line 35: OK so on review I would recommend updating the catch statement to the following:

```
		if (err instanceof Error) {
			setStatus('db', { status: Status.ERROR, info: { message: err.message } });
		} else {
			const error = String(err);
			setStatus('db', { status: Status.ERROR, info: { message: error } });
		}
```

Reference: https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#catch_binding

This way we can insure error message is displayed in Swagger / Health check, and we can update the `info` type accordingly

apps/api/src/server.ts
```diff
@@ -20,16 +20,44 @@
...
+	app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(yaml.load(path.join(__dirname, './resources/swagger.yaml'))));
+
 	app.get('/', (_req: Request, res: Response) => {
 		res.send('Hello World!');
 	});
 
+	app.get('/health', (_req: Request, res: Response) => {
+		const health = getHealth();
+
+		// TODO: implement proper versioning logic when in place
```
demariadaniel, line 47: Good suggestion, could leverage `const { npm_package_version } = process.env;` then replace V1 in resbody.version

demariadaniel (approved):
Looks great!! Key piece of the server up and running

Issue #63, opened 2024-11-19
Revoke and Withdraw


Buwujiu:
Add a mechanism to support 3 paths:
1. DAC approved -> revoked,
2. Rep Review -> Draft
3. DAC Review -> Draft

demariadaniel:
Don't see a ticket for 'Closed' so I will include it here

Issue #62, opened 2024-11-19
DAC Reject


Buwujiu:
Implement DB updates for the two Rejection state transitions:
- DAC Review -> DAC reject
- Institutional Rep Review -> institutional Rep reject

- [ ] The following DB updates occur as a single transaction
  - [ ] Add application action to DB for `dac_review_reject` or `institutional_rep_reject`
  - [ ] Update state of application in DB

Buwujiu:
PR is merged, closing.

Issue #61, opened 2024-11-19
DAC Revision Path


Buwujiu:
Add a mechanism to support path: 
DAC revisions -> applicant submit revisions

Issue #60, opened 2024-11-19
Revision Path


Buwujiu:
Add a mechanism to support path: 
Draft -> rep revisions -> submit revisions -> ,

PR #86, opened 2024-12-05
#58 Feat: Setup State Transition management

demariadaniel:
### Related Issues

- #3 
- [#58](https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=87879785&issue=Pan-Canadian-Genome-Library%7Cdaco%7C58)
- #59 
- #60 
- #61 
- #62 
- #63 

### API
- Implements [Typescript FSM](https://github.com/eram/typescript-fsm)-based Class for managing Application State transitions
- Class methods are added for every transition, with private post-transition callbacks for each event
- Unit tests for every method
- Adds State diagram to Docs

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [ ] Documentation


apps/api/src/api/states.ts
```diff
@@ -0,0 +1,292 @@
...
+			} else {
+				return failure(`Cannot submit application with state ${this.getState()}`);
+			}
+		} else {
+			return failure(`Cannot submit application with state ${this.getState()}`);
+		}
+	}
+
+	async submitRepReview() {
+		return this.submitDraft();
```
demariadaniel, line 84: Stubs to represent each transition. These will be updated with additional validation logic in the future. Currently they just use a base function to validate state transition is allowed, then dispatch the update.

apps/api/src/api/states.ts
```diff
@@ -0,0 +1,292 @@
...
+	async submitRepRevision() {
+		return this.submitDraft();
+	}
+
+	async submitDacRevision() {
+		return this.submitDraft();
+	}
+
+	private async _onSubmit() {
+		return success('post dispatch on submit');
```
demariadaniel, line 96: Example post-transition callback. Likely place for DB updates. Requirements TBD.

apps/api/src/api/states.ts
```diff
@@ -0,0 +1,292 @@
...
+
+	// Handler Methods
+	// Submit
+	async submitDraft() {
+		if (this.can(submit)) {
+			// TODO: Add Validation
+			const validationResult = await validateContent(this._application);
+			if (validationResult.success) {
+				await this.dispatch(submit);
+				return validationResult;
```
demariadaniel, line 74: Example public method. Validates state transition and related record before allowing event.

apps/api/src/api/states.ts
```diff
@@ -0,0 +1,292 @@
...
+type ApplicationTransitions = ITransition<
+	ApplicationStateValues,
+	ApplicationStateEvents,
+	ApplicationTransitionCallback
+>;
+
+export class ApplicationStateManager extends StateMachine<ApplicationStateValues, ApplicationStateEvents> {
+	private readonly _id: number;
+	private readonly _application: typeof applications.$inferSelect;
+	private readonly _initState: ApplicationStateValues;
```
demariadaniel, line 64: _initState is not used for anything currently so not sure if we truly need it. Maybe to implement a `revert` method.

apps/api/tests/api/application-state.test.ts
```diff
@@ -0,0 +1,211 @@
...
+			assert.ok(result.success);
+
+			testStateManager = result.data;
+			assert.ok(testStateManager);
+		});
+
+		it('should initialize with state DRAFT', () => {
+			stateValue = testStateManager.getState();
+			assert.strictEqual(stateValue, DRAFT);
+		});
```
demariadaniel, line 74: 'Happy path' tests are reusing the same `testStateManager` and updating it from one test to the next. If preferred I can update this to create a new Class on every test, so each test is standalone.

docs/states/DACO State Diagram.png
demariadaniel, line 1: Could potentially add more State Machine documentation

apps/api/src/api/states.ts
demariadaniel: Looking at organization, I'm wondering if I should move private methods to the top (because they're used in the constructor) then public methods to the bottom (convention to have the constructor at the top, public methods used after instantiation, etc)

JamesTLopez (commented):
Just one request from my part but lgtm 🔥

apps/api/src/api/states.ts
```diff
@@ -0,0 +1,292 @@
...
+
+type ApplicationTransitions = ITransition<
+	ApplicationStateValues,
+	ApplicationStateEvents,
+	ApplicationTransitionCallback
+>;
+
+export class ApplicationStateManager extends StateMachine<ApplicationStateValues, ApplicationStateEvents> {
+	private readonly _id: number;
+	private readonly _application: typeof applications.$inferSelect;
```
joneubank, line 63: Should this be readonly? after a state transition the application content will be updated, no?

docs/states/DACO State Diagram.png
joneubank, line 1: 🙏

apps/api/src/api/states.ts
```diff
@@ -0,0 +1,292 @@
...
+	const result = await service.getApplicationById({ id });
+	if (!result.success) {
+		return result;
+	}
+
+	const dbRecord = result.data;
+	const appStateManager = new ApplicationStateManager(dbRecord);
+
+	return success(appStateManager);
+};
```
joneubank, line 292: Could we instead change this function to take the application as the input? A pattern I expect to see is that an application is read from the DB, some action is taken from it, and then a state transition is required. So would at that time want to create the state manager from the application and perform the action.

apps/api/src/api/validation.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { applications } from '../db/schemas/applications.js';
+import { AsyncResult, success } from '../utils/results.js';
+
+// TODO: Add Validation
+export const validateContent = async (
+	application: typeof applications.$inferSelect,
+): AsyncResult<typeof applications.$inferSelect> => {
+	return success(application);
+};
```
joneubank, line 28: Wow, so valid!

apps/api/src/api/states.ts
joneubank: Alternately, we could have both `createApplicationStateManager(application)` and `getApplicationWithStateManager(id: number)` as a convenience to do both steps at once.

apps/api/src/api/states.ts
joneubank: The application-servcie could export `type ApplicationData = typeof applications.$inferSelect;` for use around the rest of this app.

joneubank (dismissed):
I would recommend the change proposed to the exported function, but generally I think this looks nice to work with.

apps/api/src/api/states.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import { ApplicationStates, ApplicationStateValues } from '@pcgl-daco/data-model/src/types.js';
-import { StateMachine, t as transition } from 'typescript-fsm';
+import { ITransition, StateMachine, t as transition } from 'typescript-fsm';
 import { getDbInstance } from '../db/index.js';
 import { applications } from '../db/schemas/applications.js';
```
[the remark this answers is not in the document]
demariadaniel: All good, updated along with a few other related files

apps/api/src/api/states.ts
demariadaniel: True, removed readonly, and using the type variable reads a lot better

apps/api/src/api/states.ts
demariadaniel: Yes, in review the service and result logic should just live in the API ... but then this function just becomes `	const appStateManager = new ApplicationStateManager(dbRecord);`

So `createApplicationStateManager` basically is a convenience method, the class constructor takes the application as an argument

JamesTLopez (approved):
🔥 Very clean!

Issue #82, opened 2024-11-29
Add Device Warning Banner


Buwujiu:
Add the warning banner to all pages. So users will be notified if they are using a mobile or table. 
Mockup: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2767-12960&t=9Pv9Ghelu3f8m2ME-4

Issue #81, opened 2024-11-28
Build Homepage for Mobile and Tablet


Buwujiu:
## Descriptions
Build the homepage for mobile and tablet. 
Ignore the Login page for now. 
Mockups: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2706-12835&t=YSgPtUE4Po96Zd7z-4

PR #91, opened 2024-12-10
#81: Build homepage for mobile and tablet

surajgoraya:
## Summary

Implements responsiveness to the homepage, menu/header bar, as well as the footer. Fixes some issues with large displays (1080p and above) by limiting the max width for better readability. Finally, also adds A11Y fixes and tweaks to the UI to further match the mockup.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/81

### UI
- Added `useMinWidth.ts` as a globally accessible react hook. 
    - This is used to programatically get the current screen width, used in conjunction with `Breakpoints` to align with Ant Design's breakpoint sizes. 
    - Particularly useful for the vast majority of antd components which don't have their own built in responsive props like `Col` does.
    - Helps prepare for adding the device size warning mentioned in #82.
- On large screen sizes, the website would stretch to 90% of the screen width, leading it to become unreadable on wide monitors, or those with large resolutions (ex. 4k). This has now been limited with a `maxWidth` in `ContentWrapper`.
-  Header now behaves responsively. 
   - Utilizes the Ant Design `Drawer` component to act as the tablet and mobile menu.
   - Required storing links as a object since they need to be rendered in two different places. May be useful to further centralize this eventually as the Footer has some of the same links.
- Footer now behaves responsively. 
- Landing page now behaves responsively. 
- Refactored some of the header, footer and main page to use ant design's spacing tokens to further standardize it into ant design's UI library. Aligned the UI closer to Figma mockup.
- Fixed some quick A11Y issues (Image alt text, aria labels, etc...).
- Added "Apply for Access" modal.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation

### DIFF
--- a/apps/ui/src/components/Footer.tsx
+++ b/apps/ui/src/components/Footer.tsx
@@ -95,7 +95,7 @@ const FooterComponent = () => {
 		<ConfigProvider theme={pcglFooterTheme}>
 			<Footer style={footerStyle}>
 				<Link target="_blank" style={{ margin: minWidth <= Breakpoints.XL ? '1rem 0 0 0' : '0 -8rem 0 0' }}>
-					<Image width={200} src={PCGLFOOTER} preview={false} />
+					<Image width={200} src={PCGLFOOTER} preview={false} alt="Pan-Canadian Genome Library / Librairie Pancanadienne de Génomique" />
 				</Link>
 				<Flex style={{ ...contentWrapperStyles, width: '100%' }} flex={1} vertical gap={24}>
 					<Flex
--- a/apps/ui/src/components/Footer.tsx
+++ b/apps/ui/src/components/Footer.tsx
@@ -22,6 +22,7 @@ import { ConfigProvider, Flex, Image, Layout, Typography } from 'antd';
 import PCGLFOOTER from '@/assets/pcgl-logo-footer.png';
 import { pcglFooterTheme } from '@/components/providers/ThemeProvider';
 import { Breakpoints, useMinWidth } from '@/global/hooks/useMinWidth';
+import { contentWrapperStyles } from './layouts/ContentWrapper';
 
 const { Footer } = Layout;
 const { Text, Link } = Typography;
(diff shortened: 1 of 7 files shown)


apps/ui/src/components/Header.tsx
```diff
@@ -17,49 +17,184 @@
...
+	/**
+	 * Generates the links to display in the mobile and desktop menus.
+	 * @param menuLinks An array containing the list of links for the menu
+	 * @param position The 'side' of links you want in the menu, left (next to the logo) or right (away from the logo on the other side of the screen)
+	 * @returns JSX Element array containing the link elements.
+	 */
+	const displayMenuLinks = (menuLinks: MenuLinkType[], position: 'left' | 'right' | 'both'): JSX.Element[] => {
+		return menuLinks
+			.filter((ml) => (position !== 'both' ? ml.position === position : ml.position))
+			.map((ml, key) =>
```
demariadaniel, line 117: Generally would prefer `link` or `menuLink` over `ml`, short character names can become 'unsearchable' when a project becomes sufficiently large. Descriptive names provide useful context.

apps/ui/src/components/Header.tsx
```diff
@@ -17,49 +17,184 @@
...
+		width: isResponsiveMode ? '100%' : 'auto',
+	};
+
+	/**
+	 * Default action when a button in the menu is clicked, used particularly for the mobile menu which should close after click.
+	 * @param buttonAction The function for the action needed to be preformed.
+	 */
+	const onMenuButtonClick = (buttonAction?: VoidFunction) => {
+		if (!buttonAction) {
+			return;
```
demariadaniel, line 64: My first reaction is that if `onMenuButtonClick` instantly returns when `buttonAction` is `undefined` then this argument shouldn't be optional and I would revise the implementation of onMenuButtonClick
Just reviewing what that might look like

apps/ui/src/components/Header.tsx
demariadaniel: Also -- variable is `ml`, key is `mi-${key}`, inconsistent and I'm not sure what `mi` signifies

apps/ui/src/global/hooks/useMinWidth.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Breakpoints for common device sizes, based off of breakpoints provided by Bootstrap 4.0, which are used by Ant Design components.
+ * @see(@link https://getbootstrap.com/docs/4.0/layout/overview/#responsive-breakpoints)
+ * @see(@link https://ant.design/components/grid#design-token)
+ */
+export const Breakpoints = {
+	XL: 1200,
+	LG: 992,
+	MD: 768,
+	SM: 576,
+};
```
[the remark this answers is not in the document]
surajgoraya: Good catch! I didn't realize there were tokens for screen size, I can move this over to using that since the breakpoints should be based off of the antd ones anyway

apps/ui/src/components/Header.tsx
surajgoraya: Ah, yeah theres a bit of weirdness there that I don't like either. The main reason it's like this is because `MenuLinkType` doesn't require you to give an `onClickAction` since usually you're just making links. If there's a better way to handle this let me know! 

```TS
interface MenuLinkType {
	name: string;
	isButton?: boolean;
	buttonProps?: ButtonProps;
	onClickAction?: VoidFunction;
	href?: string;
	position: 'left' | 'right';
}
```

apps/ui/src/components/Header.tsx
surajgoraya: good catch! that was def a typo, fixed!

apps/ui/src/components/Header.tsx
demariadaniel: I would separate MenuLink into multiple types; a base MenuItem type with name and position, then a MenuLink and MenuButton type which extend MenuItem.

At first I was thinking that's overkill for one component, but come to think of it we will probably have more menus and can possibly reuse these definitions in the future.

demariadaniel: Similar thing with `href` -- every Link should have an href, but some items are buttons and some are links, so it's optional.

surajgoraya: This makes a lot of sense! I've gone ahead and refactored the code in commit 4586b8d as well as added an additional prop for anchor tag targets into `MenuLink` for more future extensibility.

apps/ui/src/components/Footer.tsx
```diff
@@ -63,42 +66,78 @@ const policiesConditionsLinks: LinkType[] = [
...
+						>
 							{policiesConditionsLinks.map((itemLink) => (
 								<Link key={itemLink.name} style={linkStyle} underline target="_blank">
 									{itemLink.name}
 								</Link>
 							))}
 						</Flex>
 					</Flex>
+					<Text style={textStyle}>
+						&copy; 2026 PCGL Data Access Compliance Office. All rights reserved. UI v1.0 - API v1.0
```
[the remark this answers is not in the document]
surajgoraya: Good call, updated in commit 8dac6d3

demariadaniel (approved):

PR #98, opened 2024-12-16
#84: Add the Create New Application endpoint

surajgoraya:
## Summary

Adds the create new application endpoint at `/applications/create/`, where a `user_id` is provided via the request body when sending the POST request. Updates the CreateApplication service function to use the failure success model. Updates and creates corresponding tests.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/84

### Server
- Adds the create new application endpoint at `/applications/create/`, where a `user_id` is provided via the request body when sending the POST request.
  - Returns a `201 Created` HTTP response on success, `500` on Failure and `400` on malformed requests. 
- Updates the `createApplication` function inside the service layer to use the Failure and Success model.
- Adds / Updates the corresponding tests to test the new API function and be compatible with the failure and success model.
## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/routes/application-router.ts
```diff
@@ -20,11 +20,41 @@
...
+	jsonParser,
+	async (request: Request<{}, {}, { userId: string }, any>, response) => {
+		const { userId } = request.body;
+
+		/**
+		 * TODO: Temporary userId check until validation/dto flow is confirmed.
+		 * Reflect changes in swagger once refactored.
+		 **/
+		if (!userId) {
+			response.send(400).send({ message: 'User ID is required.' });
```
demariadaniel, line 44: Should this first `.send` be `.status`?

apps/api/src/resources/swagger.yaml
```diff
@@ -148,8 +118,120 @@ paths:
...
+        id:
+          type: number
+          description: The unique ID of the application in the database.
+        user_id:
+          type: string
+          description: The user ID associated with the application.
+        state:
+          type: string
+          enum:
+            [DRAFT, INSTITUTIONAL_REP_REVIEW, DAC_REVIEW, DAC_REVISIONS_REQUESTED, REJECTED, APPROVED, CLOSED, REVOKED]
```
demariadaniel, line 201: Some solid Swagger details in here 👍

apps/api/src/resources/swagger.yaml
```diff
@@ -148,8 +118,120 @@ paths:
...
+              $ref: '#/components/schemas/ApplicationCreation'
+      responses:
+        '500':
+          description: 'Error creating application'
+          content:
+            application/json:
+              schema:
+                type: object
+        '400':
+          description: 'Error crating application - No User ID specified.'
```
demariadaniel, line 142: Typo `creating application`

demariadaniel (changes requested):
Duplicate `.send` will yield `Cannot set headers after they are sent` error

apps/api/src/resources/swagger.yaml
```diff
@@ -148,8 +118,120 @@ paths:
...
+        contents:
+          type: number
+          description: A ID referring to the entry with the application contents.
+
+    ApplicationCreation:
+      type: object
+      required:
+        - user_id
+      properties:
+        user_id:
```
demariadaniel, line 231: There's a mismatch here with application router which is expecting `userId` in request.body

surajgoraya: shoot, will fix that! However, i wonder if we should standardize a convention for naming this? Internally we use `user_id`, but API / outward facing we use `userId`, should we just switch over to `user_id`?

demariadaniel: So the conflict is between JSX conventions (camelCase) vs Postgres conventions (snake case)
Which is why the UI request is typically formatted camelCase `userId` vs API snake case `user_id`

Now, in our argo clinical service, there's a fair bit of aliasing and equality management going on, you can see this file as an example:
https://github.com/icgc-argo/argo-clinical/blob/develop/src/common-model/entities.ts

We will have to standardize this, and we should add a note to a planned ticket on the roadmap.
It may be more relevant when we get to the DTO / Zod validation / error handling set up.
We could get a start on it at this point, but for the scope of this PR just changes to the names should be enough.

demariadaniel (approved):
:shipit:

Issue #94, opened 2024-12-12
Implement DACO i18n


JamesTLopez:
## Description 

Going forward into development, we want to start apply translations to the text in the app.
Integrate internationalization into app using the library [i18next](https://www.i18next.com/)

## Tasks

- [x] Setup i18next to be usable across the app
- [x] Create translation files:
    - [x] English
    - [x] French
- [x] Replace text on:
    - [x] Homepage
    - [x] Dashboard
- <s>[x] Implement language switch with url</s>
    - [x] <s>Ex:`www.{domain}.com/fr/*`. should convert the webpage into french</s>
    - [x] <s>Ex:`www.{domain}.com/en/*`. should convert the webpage into french</s>
    - [x] <s>Anything not specified will default into english</s>
- [x] retire date formatter logic in utils and use intl.api to formate that dates instead

JamesTLopez:
NOTE: language switch will be implemented POST-MVP. This ticket is to setup the foundation. [Relevant](https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=88042895&issue=Pan-Canadian-Genome-Library%7Cdaco%7C70)

PR #111, opened 2024-12-17
#94: i18n integrations

JamesTLopez:
## Summary

Intergrate i18next into repository featuring french and english translations

## Description of Changes

- create translations for english and french
- <s>create url transition logic</s>
- retire dateformat util in favor or intl.api
- integrate translations into text

### Special Instructions
Before running these changes, you will need to install `react-i18next` and `i18next`:
```
pnpm i
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] Documentation


apps/ui/src/pages/App.tsx
```diff
@@ -16,20 +16,20 @@
...
-			<HeaderComponent />
-			<PCGLRoutes />
-			<FooterComponent />
-		</Layout>
+		<Routes>
+			<Route path=":lang?" element={<LanguageLayout />}>
+				<Route index element={<HomePage />} />
+				<Route path="dashboard" element={<DashboardPage />} />
+			</Route>
+		</Routes>
```
JamesTLopez, line 32: Since I moved the providers into `main.tsx` and the `Layout` , `HeaderComponent` and `FooterComponent` to the `LanguagesLayout`, it didnt make much sense to have a` routes.tsx` and `App.tsx` with one single render of just the routes. Open to suggestions or maybe im missing something here.

apps/ui/src/i18n/locale/en/enTranslations.json
```diff
@@ -16,5 +16,12 @@
 	},
 	"modal": {
 		"authorization": "For authorization, we require a valid institutional email address. This will be the email address you will use to log in to PCGL DACO and will be the email address associated with PCGL Controlled Data Access."
+	},
+	"links": {
+		"Policies": "Policies & Guidelines",
+		"HelpGuides": "Help Guides",
+		"ControlledDataUsers": "Controlled Data Users",
+		"Apply": "Apply For Access",
+		"Login": "Login"
```
JamesTLopez: I figured that its pretty early into development and there is not a lot of translations. Leaving them in one file for now seems to be the more organized instead of many files with a couple of translations

apps/ui/src/components/layouts/LanguageLayout.tsx
```diff
@@ -17,25 +17,47 @@
...
+		const languages = Object.keys(resources);
+
+		// Continue as default language
+		if (!lang) {
+			return;
+		}
+
+		// If the language is not identified, continue as default but remove the param
+		if (!languages.includes(lang)) {
+			const cleanedUrl = pathname.replace(`${lang}/`, '');
```
surajgoraya, line 45: I'm wondering if we should figure out what the actual behaviour for this should be. 

Looking at other provincial and federal bilingual sites specifically when no language identifier is specified, the website defaults to a language selection screen (see: https://www.canada.ca/ and https://www.ontario.ca/), whereas, if a unknown language is specified (e.g https://www.canada.ca/es/ or https://www.ontario.ca/es/), it results in a 404. 

I'm wondering if it'd be wise to not assume english as our "default", but rather go to a language select splash page like Canada.ca / Ontario.ca does?

surajgoraya: Side note: these websites also translate the path URL (eg. https://www.ontario.ca/fr/page/gouvernement-de-lontario vs. https://www.ontario.ca/page/government-ontario), but this might be out of scope for now

apps/ui/src/components/pages/dashboard/ApplicationStatusBar.tsx
```diff
@@ -62,18 +60,26 @@ const ApplicationStatusBar = ({ expiresAt }: ApplicationStatusBarProps) => {
...
+							{!expiresAt ? (
 								<>
 									<CheckCircleFilled style={{ color: token.colorPrimary, fontSize: 30 }} />
 									<Flex vertical>
-										<Text strong>You have access to PCGL Controlled Data</Text>
-										<Text>Expires: {formatDate(expiresAt)}</Text>
+										<Text strong>{t('dashboard.hasAccess')}</Text>
+										<Text>
+											{t('label.expires')}:
+											{t('date.intlDateTime', {
```
surajgoraya, line 70: image

Minor, but we should probs add a space between expires and the date

surajgoraya (commented):
Just a few things (one of which is more of a overall "how do we wanna design this question" rather than one directly tied to the PR). Overall PR looks super good! :)

apps/ui/src/components/pages/dashboard/cards/ApplicationCard.tsx
```diff
@@ -34,12 +34,31 @@ type ApplicationCardProps = {
...
 
 const ApplicationCard = (props: ApplicationCardProps) => {
+	const { t } = useTranslation();
 	const { id, status, createdAt, expiresAt } = props.application;
 	const { showEdit, color, showActionRequired } = getApplicationStatusProperties(status);
 	const { token } = useToken();
 	const minWidth = useMinWidth();
 	const isLowResDevice = minWidth <= token.screenLGMax;
 
+	const formateDate = (createdAt: Date, expiresAt: Date) => {
```
demariadaniel, line 44: Typo, `formate` should be `format`

apps/ui/src/components/Footer.tsx
```diff
@@ -33,43 +34,44 @@ interface LinkType {
...
-	{
-		name: 'Terms & Conditions',
-	},
-	{
-		name: 'Publication Policy',
-	},
-];
-
 const FooterComponent = () => {
+	const { t } = useTranslation();
```
demariadaniel, line 38: Kind of a headache to refactor (lots of instances) but I'd recommend changing this to `const { t: translate } = useTranslation();` and updating usage to `name: translate('links.contact')`

apps/ui/src/pages/App.tsx
demariadaniel: It's fine, having a separate routes file may become useful when we have more routes and more functionality in App.tsx. Worth doing if/when it becomes an issues but not essential right now.

apps/ui/src/components/layouts/LanguageLayout.tsx
demariadaniel: These are definitely good suggestions worth considering. I think we'd have to involve Alexis and Patrick in designing what that looks like, and draft up a new ticket. Right now there's no design for a 'select language screen' so it's a bit vague how James would implement one.

We could potentially look at, no language = default to english, unrecognized language = throw error, for the scope of this PR?

Translating path urls as well I think there's good arguments for adding, but again this expands the scope of the ticket and might be worth discussing with the broader team.

apps/ui/vite.config.ts
```diff
@@ -11,11 +11,7 @@ export default defineConfig({
 	},
 	resolve: {
 		alias: {
-			'@/assets': path.resolve(__dirname, './src/assets'),
-			'@/components': path.resolve(__dirname, './src/components'),
-			'@/pages': path.resolve(__dirname, './src/pages'),
-			'@/hooks': path.resolve(__dirname, './src/hooks'),
-			'@/global': path.resolve(__dirname, './src/global'),
+			'@': path.resolve(__dirname, './src'),
```
ciaranschutte, line 14: 😍

apps/ui/src/components/Footer.tsx
JamesTLopez: EVEN BETTER  `const { t: metamorphosiseTranslatableText } ...` 😂 

I agree, `translate` is much better! Refactoring. 😎

apps/ui/src/components/layouts/LanguageLayout.tsx
JamesTLopez: > We could potentially look at, no language = default to english, unrecognized language = throw error, for the scope of this PR?

Initially, this was my idea, recongnize `en` and `fr` then default to english if it doesn't register. But I was having a problem with how [optional segments](https://reactrouter.com/start/library/routing#optional-segments) works with the current solution. So right now, if the language is not identified, it removes the optional param via the current logic. But it only really works if the path is not at the root.

For example: `www.example.com/emg/dashboard` will translation to `www.example.com/dashboard` and default to english, but `www.example.com/emg` will persist in the url but will still render the homepage as english (Since the url is optional parameter works at the highest level in the url and accepts anything). So adding a error page is not really possible with this solution. 

I sat on this for quite sometime and didn't want to linger too long as theres many different ways to handle this. I figured a discussion for a temporary language switch solution would be more appropriate before wasting too much time.

Also thank you @surajgoraya for bringing this up, I wanted to talk about this yesterday but I was out and I missed it in my review!

demariadaniel: I think `www.example.com/emg/dashboard` should reroute to `www.example.com/en/dashboard` and `www.example.com/emg` should route to a 404 page, would that be possible?

Or, we treat anything following root that is not recognized as a 404, so both `www.example.com/emg/dashboard` and `www.example.com/emg`
So make language required

demariadaniel: Discussed offline, I'll let James fill in

Issue #65, opened 2024-11-19
My Applications - UI integration - List Apps by User


Buwujiu:
## Description
When the List Apps by User endpoint is ready(https://github.com/Pan-Canadian-Genome-Library/daco/issues/64), connect it with the application card on My Applications page, so when users navigate to the My Applications page, they only see the applications created by them. 
## Acceptance Criteria
Connect the UI page to the correct route.

PR #113, opened 2024-12-18
#65: UI Integration - Connect Listing Applications by User

surajgoraya:
## Summary

Connects the API to the UI, and now lists applications via the API instead of mock data on `/dashboard`. 

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/65

### Special Instructions
- Before running these changes, you will need to install the `cors` middleware for express - `cors` by running: `pnpm i`
- Add `IS_DEV=true` to your .env file inside of `/apps/api/`

### Server
- Adds the `cors` middleware package from express in order to allow for accessing the API via our frontend - https://github.com/expressjs/cors
   -  This was needed to allow for cross origin access to the API, explicitly whitelisted `http://localhost:5173`.
   - Only uses the `cors` middleware if `IS_DEV` is set to true in the environment file. If not, then that logic is skipped, and default CORS rules apply.

### UI
- Added `FetchClient.ts`, which is a lightweight wrapper around fetch to allow for it to take into consideration our Vite env config for the API. 
  - The the API for `FetchClient` is the exact same as `fetch`, and the export is aliased to `fetch`. All this wrapper does it take into account out API environment variable and prepends it.

- Connected dashboard to the API.
  - *Note:* This currently uses `testUser@oicr.on.ca` via `applicationMockData.ts` for a mock user ID since we don't have login functionality working yet  
  -  Added in `<LoadingApplicationCard/>` which acts as a Skelton loading placeholder while the app fetches data (for especially slow internet connections) - see: https://ant.design/components/skeleton.
  - Added basic network error handling, and temporarily used a `<Alert/>` component to display them. 
  - Fixed some UI weirdness when user only had one application open (which also kinda makes mobile look better, despite not being my intention, two birds with one stone I guess!)
- Created / moved some types for return data into `types.ts` 
  -  Modified some of them to handle whats returned by the server.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [x] Documentation


apps/api/src/server.ts
```diff
@@ -35,6 +36,14 @@ const port = process.env.PORT || 3000;
...
 	const app = express();
 
+	if (process.env.IS_DEV) {
+		app.use(
+			cors({
+				origin: 'http://localhost:5173',
+			}),
+		);
+	}
+
```
JamesTLopez, line 46: Just wondering, if we need cors in production as well? 
Also origins might be better as a variable in an ENV file. 
@demariadaniel

apps/ui/src/components/pages/dashboard/getApplicationStateProps.ts
```diff
@@ -18,85 +18,85 @@
...
+		case ApplicationStates.REJECTED:
 			return 'Rejected';
-		case ApplicationState.REVOKED:
+		case ApplicationStates.REVOKED:
 			return 'Revoked';
-		case ApplicationState.APPROVED:
+		case ApplicationStates.APPROVED:
 			return 'Approved';
-		case ApplicationState.CLOSED:
+		case ApplicationStates.CLOSED:
```
JamesTLopez: @surajgoraya Now that I think about it, do we need this since "ApplicationStateValues" is from the pcgl package?

JamesTLopez (commented):
Just a couple of things! But functionality wise LGTM! :)

apps/ui/src/components/providers/ThemeProvider.tsx
```diff
@@ -109,6 +109,16 @@ export const pcglFooterTheme: ThemeConfig = {
 	},
 };
 
+export const pcglSkeltonTheme: ThemeConfig = {
```
demariadaniel, line 112: Typo should be `pcglSkeletonTheme`

apps/ui/src/components/pages/dashboard/getApplicationStateProps.ts
surajgoraya: I actually just got rid of this, since I currently was doing it by getting values from the translation file, i thought that'd be a better way. Take a look at apps/ui/src/i18n/locale/en/enTranslations.json and Line 79 in application card https://github.com/Pan-Canadian-Genome-Library/daco/pull/113/files#diff-758fec47d1b5d45e9b28529fb1777fc98bc5023d61db44f598ad9ecd2a935d7aR79

@JamesTLopez

surajgoraya: we can refactor to use StateValues if you think thats better?

JamesTLopez: Ahh I see, I think removing it like you did was fine. We can just continue with ApplicationStateValues

JamesTLopez (approved):

Issue #89, opened 2024-12-05
Add to Service Layer - Application History


Buwujiu:
## Descriptions
Backend work for recording Application History. it's part of the transitions task. 

## Acceptance Criteria
- Implement logic to Read and Write from DB layer. 
- Add test cases to cover all cases.

Buwujiu:
Dan is working on last feedbacks, will take a few more days.

PR #133, opened 2024-12-20
Feat/89 Application History service

demariadaniel:
## Summary
- Adds Actions service w/ methods for all Actions

### Related Issues
- [#89](https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=89904062&issue=Pan-Canadian-Genome-Library%7Cdaco%7C89)

### Server
- Adds Service for creating state transition Actions, helper functions for all Actions
(see `/docs/states` for reference)
- Includes methods for GetById and ListActions with filters and sort
- Adds Tests for each method

### Special Instructions
Includes 2 migrations to update Actions definitions
App will require stopping server, running `pnpm run build`, `pnpm drizzle-kit migrate`, then restarting the server

There are additional details related to known Drizzle issues added to the API Readme

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/tests/db/index.test.ts
```diff
@@ -17,200 +17,40 @@
...
-		await initTestMigration(db);
 	});
 
 	describe('Connection', () => {
 		it('should connect successfully', () => {
-			assert.notEqual(db, undefined);
-		});
-	});
-
-	describe('Actions', () => {
```
demariadaniel, line 62: These tests were used for db setup and are essentially just testing that Drizzle is working, they can be removed for now

apps/api/drizzle/0001_expand_actions_enum.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TYPE "public"."application_states" ADD VALUE 'REP_REVISION' BEFORE 'DAC_REVIEW';--> statement-breakpoint
```
demariadaniel, line 1: Drizzle-kit migration issue needs to be documented

apps/api/src/service/utils.ts
```diff
@@ -31,3 +32,11 @@ export const sortQuery = (sort?: Array<OrderBy<ApplicationsColumnName>>) => {
...
 	return orderByArguments;
 };
+
+export const actionsQuery = (sort?: Array<OrderBy<ActionsColumnName>>) => {
+	const orderByArguments = sort
+		? sort.map((sortBy) => (sortBy.direction === 'asc' ? asc(actions[sortBy.column]) : desc(actions[sortBy.column])))
+		: [asc(actions.created_at)];
+
+	return orderByArguments;
+};
```
demariadaniel, line 42: Could make a generic SortFunction then individual helpers for each schema

apps/api/tests/services/action-service.test.ts
```diff
@@ -0,0 +1,326 @@
...
+		it('should get all actions requested by application id', async () => {
+			const actionResult = await actionService.listActions({ application_id });
+
+			assert.ok(actionResult.success && actionResult.data);
+
+			const actionRecords = actionResult.data;
+
+			assert.ok(Array.isArray(actionRecords) && actionRecords[0]);
+			assert.strictEqual(actionRecords[0].application_id, application_id);
+		});
```
demariadaniel, line 317: Need test for sorting

apps/api/tests/api/application-api.test.ts
```diff
@@ -37,9 +39,10 @@ import {
 	testUserId as user_id,
 } from '../testUtils.js';
 
-describe('Application API', () => {
+describe.only('Application API', () => {
```
ciaranschutte, line 42: will this cause all other tests in repo to be ignored?

apps/api/src/service/application-service.ts
```diff
@@ -46,23 +47,29 @@ const applicationService = (db: PostgresDb) => ({
...
 				};
 				const newAppContentsRecord = await transaction.insert(applicationContents).values(newAppContents).returning();
 				if (!newAppContentsRecord[0]) throw new Error('Application contents record is undefined');
 
+				// Create associated Actions
+				const actionService = initActionService(db);
+
+				const actionResult = await actionService.create(newApplicationRecord[0]);
+				if (!actionResult.success) throw new Error(actionResult.errors);
+
```
ciaranschutte, line 65: "init.." says to me this is something static that won't change until we restart the server.
like initialising a config from env vars
vs
some object closure state

looks to me (without knowing full context on this project) that this is a service that will be used a lot.
So with the DI pattern you've got going on here, maybe it would feel better _something_ like:
```js
import {actionService} from 'actionService'; 
const actionResult = await actionService(db).create(newApplicationRecord[0]);
```

ciaranschutte: popped the action service as a named export in that example
the annoying thing about default exports is they make renaming them trivial.

imagine ending up with `initCreateService`, `startActionService`, `actionService`, etc scattered in the project because no one looked to see how it was been imported in other files. if you use a named export, it enforces consistent naming.

joneubank:
> `pnpm drizzle-kit migrate` will throw an error that action_types enum is already defined
will this prevent the migration from running?

demariadaniel:
> > `pnpm drizzle-kit migrate` will throw an error that action_types enum is already defined
> > will this prevent the migration from running?

It will, so this bug is definitely a pain point. I'm following the issues on Github for updates. I can see if decoupling the DBML script would resolve anything, though DBML would need additional changes to work.

apps/api/src/service/application-service.ts
demariadaniel: Good points, I will revise this tomorrow. The service setup could be improved but right now has minimal dependencies, so there's not much enforcing it be built any differently other than to reference the current DB instance as needed.

demariadaniel: OK so I've updated this in: [Named service exports](https://github.com/Pan-Canadian-Genome-Library/daco/pull/133/commits/038c402e539b5bc9f4572273f8f2abe95573d54d)
[Prefer shortened names](https://github.com/Pan-Canadian-Genome-Library/daco/pull/133/commits/567659906c0f6d10b55c9d366a6aef56c8a3ab9e)

I updated applicationService as well so it's consistent from one file to the next.
Jon and I were talking about further name / dependency setup changes, this should be a step in the right direction we can build on in upcoming tickets

apps/api/tests/services/action-service.test.ts
demariadaniel: Added sorting tests for id & created_at

apps/api/src/api/application-api.ts
```diff
@@ -32,9 +33,13 @@ import { failure } from '@/utils/results.js';
...
 export const createApplication = async ({ user_id }: { user_id: string }) => {
 	const database = getDbInstance();
-	const service: ApplicationService = applicationService(database);
+	const applicationRepo: ApplicationService = applicationService(database);
+	const actionRepo: ActionService = actionService(database);
 
-	const result = await service.createApplication({ user_id });
+	const result = await applicationRepo.createApplication({ user_id });
+	if (result.success && result.data) {
+		await actionRepo.create(result.data);
```
ciaranschutte, line 41: if this `create` fails, will it break the `actions` for the this application?

phrasing it a different way, will there a missing application state when there should be an entry with the state of `ApplicationStates.DRAFT` ?

I'm not too clear on how this project uses the state machine, but I imagine there needs to be valid state all the way through or there will be errors.

demariadaniel: Action creation is grouped in a transaction with the create step so failures should be managed:
https://github.com/Pan-Canadian-Genome-Library/daco/blob/feat/89-application-history-service/apps/api/src/service/application-service.ts#L44

ciaranschutte: I see, so does it need to happen on line 41 also?

demariadaniel: No, that can be removed. I've taken this out and moved the related Create Action unit test out of API to the service test file

apps/api/drizzle/0001_expand_actions_enum.sql
demariadaniel: Added issues to Readme

demariadaniel: Also: REP_REVISION value added, missing in main

joneubank (approved):
Read through pretty quickly but in general it looks like a clean refactor. Thank you for adjusting the naming!

demariadaniel:
> Read through pretty quickly but in general it looks like a clean refactor. Thank you for adjusting the naming!
> What is the issue? Where will we document it?

See API Readme update: https://github.com/Pan-Canadian-Genome-Library/daco/pull/133/files#diff-31a92f1c67f6af74712cadf88311a65463655d4ff829634f77dfa7434ef29dc2

Documented + contains a link to the issue

Issue #130, opened 2024-12-19
Form Validation Library PoC


Buwujiu:
## Descriptions
Do a POC on the content and form validation library
James and Dan to talk to Sam and learn from her experience. 
OHCRN is using a mixture of zod and react-hook-form.
## Acceptance Criteria
Document the PoC, and make a decision on validation library. 
Dan reviewed and approved the POC.

Buwujiu:
Dan reviewed and approved the POC. Closing.

JamesTLopez:
Relevant PR's:
- https://github.com/Pan-Canadian-Genome-Library/daco/pull/146 (merging code into main)
- https://github.com/Pan-Canadian-Genome-Library/daco/pull/139 (has the discussions on chosen libraries)

PR #146, opened 2025-01-10
#130: refactor to react-query and add zod-rhf

JamesTLopez:
## Summary

Add libraries and refactor api calls getApplicationList and createNewApplication to a react-query flow.
Also create ErrorPage component

### Related Issues

- [Form Validation POC](https://github.com/Pan-Canadian-Genome-Library/daco/issues/130)

## Description of Changes
Added libraries:
- `zod` 
- `@hookform/resolvers` (for zod/rhf integration)
- `react-hook-form`
- `antd-zod` (for zod/antd integration)
- `react-query` 

Please run `pnpm i` to install the new packages when merged.
```
pnpm i
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [x] Documentation


apps/ui/docs/FormAndDataFetching.md
```diff
@@ -0,0 +1,85 @@
...
+Libraries and bindings:
+
+- [Zod validation](https://zod.dev/)
+- [React-Hook-Form](https://react-hook-form.com/)
+  - [antd-zod](https://github.com/MrBr/antd-zod) (changes zod schema into antd rules)
+- [react-query](https://tanstack.com/query/v5/docs/framework/react/overview)
+
+(Already Included)
+
+- [Antd Forms](https://ant.design/components/form)
```
demariadaniel, line 12: I think you can remove `(Already Included)` and group this with the rest of the links

apps/ui/docs/FormAndDataFetching.md
```diff
@@ -0,0 +1,85 @@
...
+
+Antd has their own custom solution for setting [rules](https://ant.design/components/form#rule)/validation parameters. Something to consider if RHF no longer required.
+
+### React-Hook-Form
+
+React-hook-form is an alternative lightweight library that allows developers to write forms in react with ease of use, no concerns with performance and flexibility
+
+- Uses hooks to manage form inputs or elements
+- Allows usage of third-party libraries form components using their `<Controller/>` element, which will be integral with Antd
+- Easily integrated with thirdpart validators such as `Zod` but using `useForm` field passing in `zodResolver(schema)`
```
demariadaniel, line 27: Typo `thirdparty`

apps/ui/docs/FormAndDataFetching.md
```diff
@@ -0,0 +1,85 @@
...
+- production level validation with lots of community support
+- integrates with react-hook-form extremely well allowing more custom validation rules
+- having a package that shares the schemas between the frontend and the backend will sync validation solutions
+
+### React-Query (data fetching)
+
+React-query is a data fetching library that makes fetching, caching and handling api state much easier for the developer.
+By passing the fetching function (using fetchApi or Axios), and passing it through react-query hooks, it gives
+the developer a set of `state` variables such as `isLoading`,`isError` and `error` without writing any custom solution of `useEffect` and `useState`
+hooks to acquire similar functionality. react-query generally doesn't force the developer to follow their patterns or paradims and can still get custom when needed.
```
demariadaniel, line 41: Typo 'paradigms'

apps/ui/src/api/useGetApplication.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ServerError } from '@/global/types';
+import { Application } from '@pcgl-daco/data-model';
+
+const useGetApplication = (id?: string | number) => {
+	const { t: translate } = useTranslation();
+
+	return useQuery<Application, ServerError>({
+		queryKey: [id],
+		queryFn: async () => {
+			const response = await fetch(`/applications/${id}`);
```
demariadaniel, line 33: A) Is `getApplicationData` planned to be updated in another ticket?
https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/apps/ui/src/pages/dashboard.tsx#L65 

B) To what extent could we make generalized query files vs. one file per endpoint?

demariadaniel: C) `useGetData` is unused in the repo, essentially `useQuery` is playing the same role, so I believe we can remove `useGetData`?

JamesTLopez: A -  Oops, I forgot about that endpoint, ill refactor that endpoint.

B -  When I worked with React-Query, it was a per file per endpoint pattern. Seeing as most endpoints will return different data, I think its safe to use one file per endpoint as we may have to do different types of logic on the clientside. 
I'm open to suggestions for design implementations here!

demariadaniel (approved):
Tested + working & feedback incorporated, thanks for getting this set up!

surajgoraya (approved):
Looks great! Can't wait for this! 🚀

apps/ui/src/pages/applications/index.tsx
```diff
@@ -58,26 +51,15 @@ const ApplicationViewer = () => {
...
-				message: data.statusCode === 404 ? translate('errors.applicationNotFound.title') : data.message,
-				errors: data.statusCode === 404 ? translate('errors.applicationNotFound.message') : data.errors,
-			});
 		}
-	}, [data, isEditMode, navigation, translate]);
+	}, [data, isEditMode, navigation]);
 
-	// Should make a component for this
-	if (!applicationData || error || loading) return <ErrorPage error={error} loading={loading} />;
+	if (!data || isError || isLoading) return <ErrorPage loading={isLoading} error={error} />;
```
ciaranschutte, line 57: `Error` and `Loading` are distinct states.

`ErrorPage` component is doing too much. It should handle just the error state and a different component can handle loading state. They're seperate components. Changing based on props is an anti pattern. 

In this case you might use props to change what type of error you're passing to display it differently.
`<ErrorPage error="FETCH_ERROR"/>`

apps/ui/src/components/pages/ErrorPage.tsx
```diff
@@ -26,7 +26,7 @@ const { Content } = Layout;
 
 type ErrorProps = {
 	loading: boolean;
-	error?: ServerError;
+	error: ServerError | null;
```
ciaranschutte, line 29: more info on the comment further down. this works better if you seperate out the components.

What's interesting here is why would an `Error` component ever not have an error?
Only example I can think of is something like a default case, but even then I think `SeverError` should cover this. 
eg. maybe a union of a different errors like `type ServerError = FetchError | DefaultError`
This makes intent crystal clear, and a developer doesn't have to jump into the component to figure out why a null might be passed

apps/ui/src/api/useCreateApplication.ts
```diff
@@ -0,0 +1,72 @@
...
+
+			return await response.json();
+		},
+		onError: (error) => {
+			notification.error({
+				message: error.message,
+			});
+		},
+		onSuccess: (data) => {
+			navigation(`/application/${data.id}/intro/edit`);
```
ciaranschutte, line 67: I really liked your documentation here: https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/apps/ui/docs/FormAndDataFetching.md?plain=1#L39

Be careful because the goal was to "isolate our api fetching logic" but you can see there are UI interactions for `onError` and `onSuccess`. Where ever this hook is used, onSuccess will navigate it to a certain route, so it's quite a specialized hook that does more than just create an application.

ciaranschutte: `navigation` is the `navigate` function, active verb vs noun(?)
point is intent, i'm not too sure what `navigation()` is from reading it - but `navigate` is really clear

apps/ui/src/api/useCreateApplication.ts
```diff
@@ -0,0 +1,72 @@
...
+import { mockUserID } from '@/components/mock/applicationMockData';
+import { fetch } from '@/global/FetchClient';
+import { ServerError } from '@/global/types';
+import { Application } from '@pcgl-daco/data-model';
+
+const useCreateApplication = () => {
+	const navigation = useNavigate();
+	const { t: translate } = useTranslation();
+
+	return useMutation<Application, ServerError>({
```
ciaranschutte, line 33: abstracting `useMutation` into a function and then renaming the variable, you may as well return the variable the way you want it in the first place, it's more descriptive than just `useMutation`. `const { createApplication} = useCreateApplication()`

apps/ui/src/api/useGetApplicationList.ts
```diff
@@ -0,0 +1,55 @@
...
+
+				throw error;
+			}
+
+			return await response.json();
+		},
+	});
+};
+
+export default useGetApplicationList;
```
ciaranschutte, line 55: This is an interesting abstraction.

One very foundational aspect of React is the render loop.
Hooks, including custom function hooks, impact the component lifecycle.
This is important to remember.

The most basic time a component re-renders? State change.
So here: https://github.com/Pan-Canadian-Genome-Library/daco/pull/146/files#diff-a0f7ee977e1607a7ca8acd7d71494b33edc7c81981d99633aa5dc4f1922df78dR48

What happens if a user resizes their browser? The `useMinWidth` hook triggers.
State change. Which calls the custom `useGetApplicationList` hook which calls react query calls.

easiest way to see this quickly is to add this to your `useGetApplicationList` hook 
```ts
let hookCalled = 0;

const useGetApplicationList = (id?: string | number) => {
	console.log('useGetApplicationList called', ++hookCalled);
	....
}
```
..now resize the browser on the Dashboard page...

`useQuery` thankfully caches and does a tonne of stuff internally to prevent more network calls.
I'd guess `useTranslation` gets called again.

Something to be careful about.

PR #147, opened 2025-01-13
fix(cors): resolve cors not working on development

JamesTLopez:
## Issue
Resolve cors error on development.
Environment variable is a string, so we needed to compare if the value is string `false`.


demariadaniel (approved):
Thanks for catching this

Issue #137, opened 2025-01-06
Feedbacks & Bugs on Get Apps By User


Buwujiu:
## Feedbacks
Feedbacks for https://daco-api.pcgl-dev.cumulus.genomeinformatics.org/api-docs/#/Applications/get_applications: 
1. Add instruction to the swagger UI on how to filter by state. What are the filter states? 
2. Add instructions on how to sort. Add an example string is even better. Such as
Example:
[{"direction":"desc","column":"id"}]
3. What is the first page number? 0 or 1? Would be nice to indicate it on swagger UI. 
4. add total page, the current page to the response. Standardise the paging info in all endpoints. 
5. make sure the response filters the `Review` apps and return it to the top. 
6. Return more fields: user names, email, country. 

## Bugs:
1. when enter a negative number as pageSize, endpoint returns all applications. Expected: throw an error to indicate that pageSize cannot be negative.
![Image](https://github.com/user-attachments/assets/5335f21a-fa58-4b04-b483-0d810c62bb6a)
 

## Sorting Rules: 
Cannot sorting because sorting items are unclear, but here are the rules:
- Applications should be sorted by application #, Updated, Status.
- Sort By application # in ascending by default. 
- Sort by Updated in ascending by default.
- Sort by Status: Show the “For Review” applications first by default, then sort alphabetically.

edsu7:
Overall looks good but have a few questions remaining.

## Feedbacks
Feedbacks for https://daco-api.pcgl-dev.cumulus.genomeinformatics.org/api-docs/#/Applications/get_applications: 
> 1. Add instruction to the swagger UI on how to filter by state. What are the filter states? 

Instructions are nice and clear

> 2. Add instructions on how to sort. Add an example string is even better. Such asExample:
[{"direction":"desc","column":"id"}]

Examples are nice and clear

> 3. What is the first page number? 0 or 1? Would be nice to indicate it on swagger UI. 

Clarification on page start is nice and clear

> 4. add total page, the current page to the response. Standardise the paging info in all endpoints. 

Total page response is nice and clear

> 5. make sure the response filters the `Review` apps and return it to the top. 

@Buwujiu can applications be placed into review state? Not testable?

> 6. Return more fields: user names, email, country. 

Looks good, @Buwujiu do we distinguish between PI vs application submitter such as in the following [link](https://docs.icgc-argo.org/docs/data-access/daco/applying#users-involved-in-the-application)

## Bugs:
[...]

PR #149, opened 2025-01-14
#137: Address Feedback and Bugs with Get Apps by User

surajgoraya:
## Summary
Addresses the feedback and bugs found in #137. It also additionally readies the endpoint to be able to respond to the needs of the DAC Manage Application page. As a consequence of the feedback and bugs address it also makes the front end compatible with these changes. 

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/137

### Server
- Fixes the following bugs with the `/applications/userId={userID}` endpoint:
   - No longer allow negative numbers to be submitted to the endpoint for paging, and page size
      - This now responds with a `400` Client error when done.
      - As a consequence, this also now takes care of NaN values as well, which were previously being passed into the app
   - `sortQuery` function mistakingly did not sort when no sort was passed, this has been fixed.
- Adds the following features to the `/applications/userId={userID}` endpoint: 
  - The ability to filter by multiple `state` values by passing in an array, similar to the `sort` query param.
      - This is required by the DAC manage applications page.
  - Adds paging information in a `paging: {}` object returned alongside the applications.
     - This includes the total number of records, the current page, and the current amount requested.
- Swagger documentation has been updated to provide examples, and updated to reflect the new structure of the return of this endpoint (see below).
   - This includes restricting numbers and negative values on for page, and pageSize right on the swagger try out page.

Screenshot of SwaggarUI
 

### UI
- Made UI (`/dashboard`) compatible with the changes that these fixes bring.
  - This included updating types handled in `@global/types` to reflect what is actually returned by the endpoint.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/resources/swagger.yaml
```diff
@@ -63,22 +63,38 @@ paths:
           name: state
           schema:
             type: string
-          description: Retrieve specific application items in a certain State
+          description: |
+            Retrieve specific application items in a certain State. This may be an array to to get multiple application types. The states that may be requested, may be the following enum values: `["DRAFT", "INSTITUTIONAL_REP_REVIEW"," DAC_REVIEW", "DAC_REVISIONS_REQUESTED", "REJECTED", "APPROVED", "CLOSED", "REVOKED"]`.
```
demariadaniel, line 67: Typo extra 'to' in `array to get multiple application types`

demariadaniel:
> Just wanted to note that this PR doesn't add in the API endpoint to get the totals for each category (for displaying the total beside the filter category on the UI), I think we should likely make this a new endpoint instead of slapping it onto this one? I'm open to suggestions though!

Good call, want to write up a ticket/issue for this?

apps/api/src/routes/application-router.ts
```diff
@@ -88,21 +91,43 @@ applicationRouter.get('/applications', async (req: Request<{}, {}, {}, any>, res
 		return;
 	}
```
demariadaniel, line 92: Small organization optimization, I think we could move this userId check before any pagination checks are performed. (Before line 84)

Then there's several checks for page/pageSize truthiness which could be combined into a single check.
Then the check for page/pageSize isNan || is 0 become their own separate conditionals.
Hope that's clear the way I've described it.

demariadaniel: I'm also considering these may be useful as reusable functions for future page/pageSize validation

apps/api/src/service/application-service.ts
```diff
@@ -167,44 +172,92 @@ const applicationService = (db: PostgresDb) => ({
...
+						state.length ? inArray(applications.state, state) : undefined,
 					),
 				)
+				.leftJoin(applicationContents, eq(applications.contents, applicationContents.id))
 				.orderBy(...sortQuery(sort))
-				.offset(page * pageSize)
-				.limit(pageSize);
+				.offset(page * pageSize);
+
+			let allApplications = rawApplicationData.slice(page, pageSize).map((application) => {
```
demariadaniel, line 213: We can leverage destructuring here `.map(({applications, application_contents}) =>` and remove the repeat use of `application`

apps/api/src/service/application-service.ts
```diff
@@ -167,44 +172,92 @@ const applicationService = (db: PostgresDb) => ({
...
+				},
+			};
+
+			return success(applicationsList);
 		} catch (err) {
 			const message = `Error at listApplications with user_id: ${user_id} state: ${state}`;
-			console.error(message);
-			console.error(err);
+			logger.error(message);
+			logger.error(err);
```
demariadaniel, line 260: 👍 Thank you, guess there's a fair bit of console use we'll have to replace

apps/api/src/service/application-service.ts
```diff
@@ -167,44 +172,92 @@ const applicationService = (db: PostgresDb) => ({
...
 					and(
 						user_id ? eq(applications.user_id, String(user_id)) : undefined,
-						state ? eq(applications.state, state) : undefined,
+						state.length ? inArray(applications.state, state) : undefined,
 					),
 				)
+				.leftJoin(applicationContents, eq(applications.contents, applicationContents.id))
 				.orderBy(...sortQuery(sort))
-				.offset(page * pageSize)
-				.limit(pageSize);
```
JamesTLopez, line 201: I can see why we are removing the limit here, we are pulling all the records and then counting them at the end. 

I don't think this solution is an issue now (and potentially ever for the amount of data the users working with) but its worth considering just making another db call for the total count and then just re-adding the limit back so we can skip over that slicing step. The count db call should be efficient as its only counting records.

Im okay with this solution, I doubt we are going to have users having thousands of records returning as we are not working with big data.

@surajgoraya @demariadaniel Thoughts?

apps/api/src/service/application-service.ts
```diff
@@ -167,44 +172,92 @@ const applicationService = (db: PostgresDb) => ({
...
+				allApplications = [
+					...reviewApplications,
+					...allApplications.filter((applications) => applications.state !== ApplicationStates.DAC_REVIEW),
+				];
+			}
 
-			return success(allApplications);
+			const applicationsList = {
+				applications: allApplications,
+				paging: {
```
JamesTLopez, line 249: `metadata` or `pagingMetadata` would be a stronger naming here

JamesTLopez (commented):
Pretty clean! Just a couple suggestions and a thought to explore!

apps/api/src/routes/application-router.ts
surajgoraya: Sounds good, I think I understand, I'll move this over to a utility function file maybe? If i'm understanding what you're saying correctly?

demariadaniel (changes requested):
I noticed the 'updated unit tests' item is incomplete and when I checked out the branch several unit tests are failing, can you look into fixing these?

apps/api/src/routes/application-router.ts
demariadaniel: Sure, we can start utils for routes!

surajgoraya:
Note, I've added a 

> > Just wanted to note that this PR doesn't add in the API endpoint to get the totals for each category (for displaying the total beside the filter category on the UI), I think we should likely make this a new endpoint instead of slapping it onto this one? I'm open to suggestions though!
> 
> Good call, want to write up a ticket/issue for this?

I've gone ahead and created an issue for this #154 and assigned myself to it!

apps/api/src/service/application-service.ts
demariadaniel: Yes a db `count` should be efficient and then we can use `limit` instead of `slice`. Then we only grab exactly the records we need, rather than grabbing more than we need and only returning a few of them. 
Probably makes negligible difference at this time but would have an impact as the application records grow in size.

I do actually have a question about the `.slice` arguments though. I'm not sure that `.slice(page, pageSize +1)` will return the correct indexes.

So let's say there's 30 results in the db. `page` is 2 and `pageSize` is 5. 
DB query with `offset` of 10 and no limit will return items 11 - 30.
Then, `.slice(2, 6)` will return items 13 - 18. The first argument for slice is the starting index, so if you supply the page value, it will skip over the first few items.
As well, the resulting array will be 6 items long, but the page size is 5.
So I think you want `slice(0, pageSize)`

demariadaniel: If you create 20 items you can test this:

Screenshot 2025-01-16 at 5 15 45 PM

surajgoraya: These changes have been made! I'm now using SQL for counts, filtering and limiting again, please take a look whenever you all have a chance!

JamesTLopez (approved):
Looks good! Thanks for making the changes! 🔥

apps/api/src/service/application-service.ts
demariadaniel: OK! This is definitely working better but there is still one revision to make.

Given >15 records with sequential numeric ids, if I request pageSize of 5, and page 1, I see records starting at id 6, which is accurate if page index starts at 0.
But if I request page 0, we get the error stating that page must be a positive integer.

So page needs to be adjusted so that page 0 is reachable.

surajgoraya: Whoops, sorry not sure how I messed that up! I think it got messed up when I moved over that function to the utils file. fixing!

demariadaniel (approved):

Issue #150, opened 2025-01-14
Bug: Dev API URL builds incorrectly


demariadaniel:
Requests to Dev API are failing, with `undefined` added to URLs
`https://daco-ui.pcgl-dev.cumulus.genomeinformatics.org/undefined/applications?userId=testUser@oicr.on.ca`

These requests are managed by fetch client & use env variable `import.meta.env.VITE_APPLICATION_API_URL`
https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/apps/ui/src/global/FetchClient.ts

Pod Env variable is set, but does not appear in browser as global variable:
<img width="708" alt="Image" src="https://github.com/user-attachments/assets/52063f80-8628-417b-9d99-cc17b432dd01" />

JamesTLopez:
### Investigation-2

Description and solution taken from PR from @joneubank :
> The UI is statically built, it therefore takes no runtime arguments that allow us to configure the host of the API for each domain.
> The solution is to configure the system to have the API be accessible through the same domain as the UI is hosted at. This is accomplished by setting the UI's web server to proxy all requests to a specific path, /api/*, through to the API. The UI now does not need to configure a domain for the API, as it will use a relative URL on its own host, and prefix any requests with the /api path.

PR can be found at [#158](https://github.com/Pan-Canadian-Genome-Library/daco/pull/158) for more information.

PR #158, opened 2025-01-20
Connect UI to API through /api proxy

joneubank:
## Summary

This PR addresses the challenge of configuring the UI to connect to the API. This was impossible in the previous setup since the UI is statically built, it therefore takes no runtime arguments that allow us to configure the host of the API for each domain.

The solution is to configure the system to have the API be accessible through the same domain as the UI is hosted at. This is accomplished by setting the UI's web server to proxy all requests to a specific path, `/api/*`, through to the API. The UI now does not need to configure a domain for the API, as it will use a relative URL on its own host, and prefix any requests with the `/api` path.

### Related Issues

This will provide a fix to this development space deployment issue:
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/150

## Description of Changes
General description of the changes in your PR and the functionality it adds.

### UI
- Adds proxy to Vite Development server
- Development server proxy is configurable through env variable. This variable won't affect dockerfile builds, only affects local development.
- Updates the FetchClient.ts to use relative URL and prefix the proxy path.

### Dockerfile
- DACO UI now imports the nginx.conf.template file instead of a `nginx.conf` file. When nginx starts, it will use this  file to generate the proxy to the DACO API based on the environment variable `API_URL`

### Special Instructions
Developers will need to update the `/.env` file in the DACO UI application at `apps/ui/.env` to add the `API_URL` variable. Please reference the `.env.schema` file.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [x] Documentation


apps/api/docker-compose.yml
```diff
@@ -3,6 +3,7 @@ name: pcgl-daco-database
 services:
   postgres:
     image: postgres:15-alpine
+    container_name: pcgl-daco/postgres
```
demariadaniel, line 6: Good call to add the container_name but attempting to start the container throws an error:
`Error response from daemon: Invalid container name (c18b95f7c79a_pcgl-daco/postgres), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed`
You'll have to replace the `/`

apps/ui/.env.schema
```diff
@@ -1 +1,2 @@
-VITE_APPLICATION_API_URL=http://localhost:3000
+API_URL=http://localhost:3001
+PORT=3000
```
demariadaniel, line 2: `/apps/api/.env.schema` needs a similar update; API is set to use port 3000 so the two apps will need to run on different ports

apps/api/docker-compose.yml
joneubank: oops, thanks.

apps/ui/.env.schema
joneubank: Simplified my changes, I no longer fix the UI port in development. This env variable is removed, all ports for local dev will appear the same as they were before.

docs/ui-api-connection.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Managing Configuration
+
+For the DACO UI, there is a single environment variable that needs to be provided, that is the URL to use to proxy requests to the API: `API_URL`.
+
+In development, this variable is set in the `/apps/ui/.env` file, as described in the [`/apps/ui/.env.schema`](../apps/ui/.env.schema). The default value `http://localhost:3001` will connect to the DACO API when running locally with the default configuration.
+
+This variable is not required when building the DACO UI Docker image.
+
+When running the DACO UI Docker image, the `API_URL` variable needs to be provided as an environment variable. When the Docker container starts, this variable will be used to update the nginx configuation to proxy requests to wherever the DACO API is hosted.
```
demariadaniel, line 21: Typo `configuration`

docs/ui-api-connection.md
```diff
@@ -0,0 +1,41 @@
...
+
+### The Proxy Path
+
+Requests made to the UI host that start with `/api` will be passed through the the DACO API server. This path is hardcoded into the application since there was no need to modify it between environments, and there was not a clean way to manage it from a single variable.
+
+This value is hardcoded in two places:
+
+   1. [`vite.conf.ts`](../apps/ui/vite.config.ts) - This value is set as a `const` near the top of the Vite config file. It is used to setup the Vite server proxy, and it is defined as a global variable through the Vite config. This global variable is used by the data fetching code to definine the path to use for API requests. _This value will affect all deployments - it is used by the Vite dev server and is baked into the code at build time._
```
demariadaniel, line 29: Typo `code to define the path`

docs/ui-api-connection.md
```diff
@@ -0,0 +1,41 @@
...
+
+The DACO system has been designed to be built into a Docker images which will be deployed in a container orchestration system (kubernetes). In this setup, the app is built once, when the Docker image is created, and then that image ia deployed into several environments through a container management system. Running the build at deploy time would significantly complicate this process. In order to "build at deploy time", we could generate a new Docker image for every deployment, or have the application build script run inside the container when it starts. Neither of these are ideal setups. Both make deployment significantly slower as we would wait for the build to complete pre-deployment. If we include the build script in the image to run at startup time, then we will significantly increase the image's file size due to included dependencies. If we don't include dependencies in the image then builds could be flakey since they would rely on fetching dependencies from the web.
```
demariadaniel, line 39: Typo `and then that image is deployed into several...`

demariadaniel (approved):
Recent pushes fix the branch so that it runs locally, so I think we should get this deployed and tested in Dev!

There's a few readme typos, don't need to hold up merging, or I can re-approve after they're fixed

joneubank:
Thanks for review Dan. I am going to merge this so we can test it in dev. I want to make significant additions/changes to that docs file anyways, so I'll address the spelling and gramma issues you found along with those other changes.

Issue #157, opened 2025-01-16
Homepage and My Applications Page Feedback


Buwujiu:
Patrick tested homepage and My Applications page in dev, and reported a few feedback in:
https://docs.google.com/document/d/1fiRfRZARJY3-IvK7k192x-RBWmca9V7To-xszaY0wqg/edit?usp=sharing, please review them and make changes accordingly.

patrickdsantos:
Changes look great! Thanks team

PR #160, opened 2025-01-20
#157: Homepage and My Applications Page Enhancements

JamesTLopez:
## Summary

Various UI enhancements for spacing, aligning and formatting elements on the Homepage and MyApplications Page

### UI
- Increase margin for hero copy on the homepage and match alignment with secondary content
- Increase padding between the two columns in the secondary content area on homepage
- reduce the text size on the footer to 12px
- added temporary logic for displaying the `My Applications and Logout` buttons to show on pages that are not the home page. NOTE: This is temporary logic, solely for QA purpose 
- align the PCGL logo on the footer to the left on tablet view

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [x] Documentation


apps/ui/src/pages/index.tsx
```diff
@@ -72,7 +67,7 @@ const HomePage = () => {
 					</Col>
 				</Row>
 			</Row>
-			<Row style={{ ...heroStyle, width: isResponsiveMode ? '95%' : '90%' }} align={'top'} gutter={token.paddingXL}>
+			<Row style={{ ...heroStyle, width: isResponsiveMode ? '95%' : '90%' }} align={'top'} gutter={[90, 60]}>
```
surajgoraya, line 70: These changes look good (and make sense)! I'm just wondering if there are any tokens from antd that we can use or if we can make some custom ones so that we can reuse this spacing in other places as a standard thing?

JamesTLopez: Initially, that was I wanted to do, but using a token we would overwrite one of the alias token like `paddingXL`. This is the largest it can go (`the feedback wanted even more spacing`). I could override it, but then the rest of `paddingXL` would be changed which would effect `Header, SectionWrapper, dashboard` etc... I could make a custom theme for the homepage, but imo that seems like a stretch for just adding more spacing between the two content elements for just the HomePage specifically. 

We could also do something like `pcglColor` and just export it from there like you suggested! My question is, where would these custom spacing elements be used in other parts of our app ?

Open to suggestions here! Maybe theres a token thats appropriate for this use-case. This is a style really only for this element so I didn't think too much of it.

JamesTLopez: Also to keep in mind, `90` doesn't mean `90px` left and right. It's actually `45px` left and right. So if i change a token, it would be WAY bigger than the previous margin/padding option: 

image

surajgoraya: > Open to suggestions here! Maybe theres a token thats appropriate for this use-case. This is a style really only for this element so I didn't think too much of it.

Hmm, I guess the homepage is a special case anyway, so yeah maybe we can forego the whole standardized token thing!

surajgoraya:
Some other bugs I caught sorry.

On the homepage, the "Get Started" button is misaligned:

Homepage

Mobile view - The logo is left aligned, whereas it needs to be centred (see mockups).
image

Tablet View - The footers in the mockup need three columns, I missed this the first time I did the responsive thing too lol, it might requires some work to get this to work tbh.
image

Mockup for reference:
image

apps/ui/src/pages/index.tsx
JamesTLopez: @demariadaniel What do you think? I'm pretty split on this one.

apps/ui/src/pages/index.tsx
```diff
@@ -58,7 +58,7 @@ const HomePage = () => {
 						<Flex vertical>
 							<Title style={{ color: token.colorTextSecondary }}> {translate('homepage.title')}</Title>
 							<Paragraph style={{ color: token.colorTextSecondary }}>{translate('homepage.introduction')}</Paragraph>
-							<Col span={6}>
+							<Col span={6} style={{ padding: 0 }}>
```
JamesTLopez, line 61: So the gutter property applied by the parent `<Row/>` actually applies to grand children as well. This is just to prevent the styling from applying on the grand children

apps/ui/src/pages/index.tsx
demariadaniel: Always good to look for ways to standardize + add reusability but in this case these values are only used in this file

apps/ui/src/components/Footer.tsx
```diff
@@ -75,12 +76,10 @@ const FooterComponent = () => {
 	const linkStyle: React.CSSProperties = {
 		textAlign: 'center',
 		textWrap: 'nowrap',
-		fontSize: minWidth <= token.screenLG ? token.fontSize : token.fontSizeLG,
 	};
 
-	const textStyle: React.CSSProperties = {
+	const complicanceTextStyle: React.CSSProperties = {
```
demariadaniel, line 81: Typo this should be `complianceTextStyle`

apps/ui/src/components/Footer.tsx
```diff
@@ -137,7 +144,7 @@ const FooterComponent = () => {
 							))}
 						</Flex>
 					</Flex>
-					<Text style={textStyle}>
+					<Text style={complianceTextStyle}>
 						&copy; {new Date().getFullYear()} PCGL Data Access Compliance Office. All rights reserved. UI v1.0 - API
 						v1.0
 					</Text>
```
demariadaniel, line 150: Actually I have a question about the Footer link columns, I am seeing this two-column layout when I have Chrome set to various responsive dimensions:
Screenshot 2025-01-22 at 11 40 38 AM

In the feedback doc item #5 is to update the Footer to use 3 columns as in the designs:
https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2767-12707&t=c7SOns0CM4i9tY9U-0

So this might need a second look

JamesTLopez: Okay! I made the changes, I had to do some decent refactoring in that Footer element and was a lot more work than I previously thought. 

The copyright and compliance text exists in a different element entirely for desktop views vs tablet/mobile so i had to do some breakpoint conditional checks to render different elements. If we don't like this solution, I did run it by patrick to just simply add the `privacy policy, Terms and conditions , Publication Policy` links to the rest of the links on the desktop view, that way we don't have to rely on JS as much. That can be plan B.

demariadaniel (approved):
LGTM! 👍

Issue #138, opened 2025-01-08
Feedbacks on Get App By ID 


Buwujiu:
## The Issue
the get Apps endpoint - https://daco-api.pcgl-dev.cumulus.genomeinformatics.org/api-docs/#/Applications/get_applications: the returned application object has these info:
```
{
    "id": 1,
    "user_id": "900",
    "state": "DRAFT",
    "createdAt": "2025-01-06T20:30:12.241Z",
    "updatedAt": null
  },
```
but the returned object from https://daco-api.pcgl-dev.cumulus.genomeinformatics.org/api-docs/#/Applications/get_applications__applicationId_ returns more info:
```
{
  "id": 3,
  "user_id": "900",
  "state": "DRAFT",
  "created_at": "2025-01-06T20:32:23.725Z",
  "approved_at": null,
  "updated_at": null,
  "expires_at": null,
  "contents": 3
}
```

## Feedbacks
2.  update the property alias in get by ID endpoint to use camel case.
3. `contents` currently returns a number, update it so that it returns the actual application details object.

edsu7:
Feedbacks have been addressed. Please see below for suggestion and details.

> update the property alias in get by ID endpoint to use camel case.

Properties now are camel case

> contents currently returns a number, update it so that it returns the actual application details object.

Contents object contains info rather than a number.

Suggestion:
Harmonize property fields between endpoints e.g. `firstName` vs `applicantFirstName`
in `get/applications`
```
      "id": 1,
      "userId": "900",
      "state": "DRAFT",
      "createdAt": "2025-01-06T20:30:12.241Z",
      "updatedAt": "2025-01-15T15:52:36.360Z",
      "applicant": {
        "createdAt": "2025-01-06T20:30:12.247Z",
        "firstName": "Alexis",
        "lastName": "Cooper",
        "email": "ali@oicr.on.ca",
        "country": "string",
        "institution": "{\"hello\",\"goodbye\"}"
      }
    },
```
in `get/applications/{applicationsId}`
```
{
  "id": 1,
  "userId": "900",
  "state": "DRAFT",
  "createdAt": "2025-01-06T20:30:12.241Z",
  "approvedAt": null,
  "updatedAt": "2025-01-15T15:52:36.360Z",
  "expiresAt": null,
  "contents": {
    "applicationId": 1,
    "createdAt": "2025-01-06T20:30:12.247Z",
    "updatedAt": "2025-01-15T15:52:36.360Z",
[...]

PR #171, opened 2025-01-24
🐞 Bug / #138 Feedback on Get App by ID

demariadaniel:
## Summary

Adds camelCase keys and complete `contents` object to getAppById endpoint

### Related Issues

[#138](https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=93002911&issue=Pan-Canadian-Genome-Library%7Cdaco%7C138)

## Description of Changes
- Use `getApplicationWithContents` for `getAppById` endpoint to retrieve contents with application
- Add utility function to convert snake case keys to camel case keys
- Updated Swagger accordingly
- Update `@data-model` types to reflect current api response for ui
- Remove demo data from `@data-model`
- Remove fields for Ethics Data but include Requested Studies (discussed offline)

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/api/application-api.ts
```diff
@@ -102,7 +103,12 @@ export const getApplicationById = async ({ applicationId }: { applicationId: num
 	const database = getDbInstance();
 	const applicationRepo: ApplicationService = applicationService(database);
 
-	const result = await applicationRepo.getApplicationById({ id: applicationId });
```
demariadaniel, line 105: Just changing the service method used will accomplish the goals of the ticket, without affecting existing usage of getApplicationById

apps/api/src/service/applicationService.ts
```diff
@@ -154,15 +156,15 @@ const applicationService = (db: PostgresDb) => ({
...
+	getApplicationWithContents: async ({ id }: { id: number }): AsyncResult<JoinedApplicationRecord> => {
 		try {
 			const applicationRecord = await db
 				.select()
 				.from(applications)
 				.where(eq(applications.id, id))
 				.leftJoin(applicationContents, eq(applications.contents, applicationContents.id));
 
-			if (!applicationRecord[0]) throw new Error('Application record not found');
+			if (!applicationRecord[0]) throw new Error('Application record is undefined');
```
surajgoraya, line 167: Just double checking that the API router is catching this error? We're (unfortunately) relying on stiring compares right now so I want to makes sure

packages/data-model/src/types.ts
```diff
@@ -158,76 +209,31 @@ export type RevisionRequest = {
...
 
-export type Files = {
-	id: number; // TODO: Implement BigInt;
-	applicationId: number; // TODO: Implement BigInt;
+export interface Files {
+	id: number;
+	applicationId: number;
 	type: FileType;
-	SubmitterUserId: number; // TODO: Implement BigInt;
+	SubmitterUserId: number;
```
JamesTLopez, line 235: Can we camelCase here?

packages/data-model/src/types.ts
```diff
@@ -97,53 +97,104 @@ export const ApplicationMessages = {
...
-	title: string;
-	firstName: string;
-	middleName: string;
-	lastName: string;
-	suffix: string;
-	primaryAffiliation: string;
-	institutionalEmail: string;
-	researcherProfileURL: string;
-	positionTitle: string;
+export interface Applicant {
```
JamesTLopez, line 101: Pretty minor and up to you if you want to change. Maybe adding `Type` to the end of these helps differentiate the functionality's ? 
For example `ApplicationService` is a type and `applicationService` is the actual service and they are only differentiated by its casing. 

Unless, ofc theres a standard here in ts-land that i'm not aware of!

JamesTLopez (changes requested):
Looks pretty 🔥  code wise and easy to read. Can you add the image of the test's passing and adjust the camelCase issue? Then its good to go imo!

packages/data-model/src/types.ts
demariadaniel: Yep, updated this and `submitted_at`, thanks

packages/data-model/src/types.ts
demariadaniel: Yes I agree, there could be better separation between the various type/variable names.

The only 'standard' I can think of is, I wouldn't necessarily add `Type` to every type variable name either. 
For a first pass that will work: 
But if the type name is sufficiently descriptive and different from the variable, then the casing is enough to signify it's a Type. (plus using `import { type TypeName }`)

I would gravitate more toward something like `ApplicationModel` (database model), `ApplicationDTO` (UI / request object), `ApplicationService` where `/applicationService.ts` exports maybe `{ applicationSvc }` (modify export name)

"There are only two hard things in Computer Science: cache invalidation and naming things."

What do you think?

demariadaniel:
Test run successful:
Screenshot 2025-01-28 at 2 23 16 PM

packages/data-model/src/types.ts
JamesTLopez: Naming is indeed hard XD

> But if the type name is sufficiently descriptive and different from the variable, then the casing is enough to signify it's a Type. (plus using import { type TypeName })

Im agree with this! The current convention wasn't outright confusing, so i'm fine with this going forward! Thanks for taking a look!

apps/api/src/service/applicationService.ts
demariadaniel: Very good question, so `getById` is fine but I've updated `/edit` and also found that `approve` needed a check for this:
https://github.com/Pan-Canadian-Genome-Library/daco/pull/171/commits/2c7f9c462fe87a39b7310d4946febbbaa5986920

packages/data-model/src/types.ts
demariadaniel: Updated as we discussed: https://github.com/Pan-Canadian-Genome-Library/daco/pull/171/commits/115e5a0c38a6ca84cdbd2ddb8bc00ddee1f509b4

Tests passing and packages building:
Screenshot 2025-01-28 at 5 07 31 PM

JamesTLopez (approved):
🔥

demariadaniel:
> 🔥

Hurray, thanks!

Issue #154, opened 2025-01-15
Create API Route to Get Application State Totals


surajgoraya:
## Descriptions

As discussed in the PCGL Planning Meeting and in #137 (see here: https://github.com/Pan-Canadian-Genome-Library/daco/pull/149#issuecomment-2593121245), it's likely required that the DACO Manage Applications page (see #128) be able to request how many applications in total are in each category to display its filters. 

## Acceptance Criteria
* A API call which returns a JSON object containing the totals for all the various application states.

Issue #128, opened 2024-12-19
Manage Application - API Integration


Buwujiu:
## Descriptions
Connect the Get applications(https://github.com/Pan-Canadian-Genome-Library/daco/issues/64) endpoint with UI, so all applications can be found on the Manage Application page.

## Acceptance Criteria
1. refer to spec: https://docs.google.com/document/d/1NK5j-ZgwPDwkUmMvC3KfNIdXAhm8JrWmSWBekOtN_SM/edit?tab=t.0#heading=h.vcb7h43dgdus
2. Add a summary section to the get apps endpoints, the summary section is mostly for DACs. 

## Note:
Since we don't have the auth layer right now, this page will be accessible to all users.

surajgoraya:
Issue closed due to GitHub automation - reopening.

Buwujiu:
Tested in dev, found 1 bug: 
The total number of applications is 26, but the page only shows 14 apps. 
Expected behaviour: DAC should see all applications on the manage applications page.

surajgoraya:
Bug on Dev likely exists with the app metadata object on the endpoint, see: `applicationRecordsCount` - filtering by user id, whereas DACO members should see everything. Waiting for Jon's PR (#221) to proceed.

Buwujiu:
Need to change role to DAC in order to test, moving it to next sprint.

Buwujiu:
Tested in QA, looks great! Closing!

PR #174, opened 2025-01-27
#128: Manage Application Page - API Integration

surajgoraya:
## Summary
A long time coming! This PR integrates the `/applications/` API endpoint into the DACO Manage Applications page including refactoring the existing UI to handle all sorting / pagination via the backend ＼(＾O＾)／

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/128 (Primary)
   - https://github.com/Pan-Canadian-Genome-Library/daco/issues/154 (Related)
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/137 (Related)

### Special Instructions
Before running these changes, you will need to re-build the `@pcgl-daco/data-model` package to pick up the new type guard util function:
```
# from root folder
pnpm build:all
```

### UI
- Splits the structure of the frontend manage applications page to have a `index.tsx` as a primary entry point under `/pages` and a `ManagementDashboard.tsx` under `/components` splitting up logic. This follows the pattern used in other new frontend pages.
- Refactors `useGetApplicationList.ts` to support all sorting, paging and filtering options.
- Refactors UI to place state into the URL to allow state to be shared via link
   - _NOTE:_ `antd` table does not allow for modification of the sort state programatically, so sorting is not stored in the URL bar, however, both application status / state filters and pagination are stored and can be shared via link.
   - When page loads, it checks URL state to determine if the filters provided is valid, if not it'll default back to TOTAL and page 1.
- Adds `useGetApplicationCounts.ts` under `/api/metadata` to fetch application type counts.
   - Updates related typing information.
- Adds `apps/ui/src/global/utils.ts` - which contains a utility to determine if a page number is valid, similar to the one found on the server.
- Adds more generic HTTP status error copy in `enTranslations.json` to be used when an error occurs with the API.
- Fixes:
   - Broken references to `REP_REVISON` replacing them with `INSTITUTIONAL_REP_REVISION_REQUESTED`
   - Broken type information for API in `apps/ui/src/global/types.ts`
- Chores:
   - Renames  StatusTableColumn --> ApplicationStatusColumn.tsx for clarity.
   - Renames some type variables used on the frontend for clarity.

### Server
- Fixes:
   - API not always returning `DAC_REVIEW` applications first if more than 20 records by relying on SQL orderby defaults.
      - Tests for `updated_at` and `created_at` updated to handle this change since they expected the API to always return  `created_at` ordered by asc if nothing was passed.
   - Server crashing if a malformed JSON object was passed to `state` or `sort` filters.
      - This will throw a 400 error now, `swagger.yaml` has been updated. 
  
## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
  
Test Results

- [x] Documentation


apps/ui/src/pages/manage/applications/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+			return 'approved_at';
+		case 'updatedAt':
+			return 'updated_at';
+		case 'expiresAt':
+			return 'expires_at';
+		case 'state':
+			return 'state';
+		default:
+			return 'id';
+	}
```
demariadaniel, line 260: I am implementing something similar in #171. This makes me wonder what will be a good reusable approach to getting names a -> b and b -> a. Something like `{ a : b} as const` in the shared `@types` package.
https://github.com/Pan-Canadian-Genome-Library/daco/pull/171/files#diff-21c9756b67e8bda64179cb191ae8ac7f5aadd31f5f8ff52955b574b94b6f34f8

We're at the point now where we can connect some of these dots

apps/ui/src/i18n/locale/en/enTranslations.json
```diff
@@ -99,6 +99,14 @@
 			"404": {
 				"title": "The requested resource could not be found.",
 				"message": "Sorry, looks like the resource you're trying to access does not exist"
+			},
+			"400": {
+				"title": "That was a bad request.",
+				"message": "Sorry, looks your browser sent a bad request, please reload the page and try again."
```
demariadaniel, line 105: Should this be "looks like your browser sent..." ? Missing `like`

surajgoraya: Whoops, thanks for catching this!

apps/ui/src/pages/manage/applications/index.tsx
surajgoraya: I think it makes sense to centralize this, the implimentation in #171 makes sense! Would you like me to wait till thats merged in and impliment this using that approch?

apps/ui/src/api/useGetApplicationList.ts
```diff
@@ -22,14 +22,41 @@ import { useTranslation } from 'react-i18next';
...
+	direction: 'desc' | 'asc';
+	column: 'user_id' | 'id' | 'created_at' | 'updated_at' | 'state' | 'approved_at' | 'expires_at';
+}
+interface ApplicationListParams {
+	userId: string;
+	state?: ApplicationStateValues[];
+	sort?: ApplicationListSortingOptions[];
+	page?: number;
+	pageSize?: number;
+}
```
demariadaniel, line 37: So I think ApplicationListSortingParams could be moved to the shared `types` package

ApplicationListSortingOptions essentially duplicates this: https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/apps/api/src/service/types.ts#L40

Main difference is the api definition allows sorting by contents but that could easily be omitted or a separate type created for 'SortingColumns'

apps/ui/src/components/pages/manage/DashboardFilter.tsx
```diff
@@ -58,12 +60,12 @@ const DashboardFilter = ({ onFilterChange, availableStates }: DashboardFilterPro
 			return {
 				name: translate(`application.states.${possibleState}`),
 				amount: availableStates.find((state) => state.key === possibleState)?.amount || 0,
-				key: possibleState as FilterKeyType,
+				key: possibleState as FilterKeys,
```
demariadaniel, line 63: This block can be changed to avoid use of `as`: 

```
		...Object.values(ApplicationStates).map((possibleState: FilterKeys): Filter => {
			return {
				name: translate(`application.states.${possibleState}`),
				amount: availableStates.find((state) => state.key === possibleState)?.amount || 0,
				key: possibleState,
			};
		}),
```

apps/ui/src/pages/manage/applications/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+import { useTranslation } from 'react-i18next';
+import { useSearchParams } from 'react-router';
+
+const { Content } = Layout;
+
+export interface TableProperties {
+	pagination: TablePaginationConfig;
+}
+
+const POSSIBLE_FILTERS = ['TOTAL', ...Object.values(ApplicationStates)] as FilterKeys[];
```
demariadaniel, line 45: Similarly, we try avoiding `as`, this should be: `const POSSIBLE_FILTERS: FilterKeys[] = ['TOTAL', ...Object.values(ApplicationStates)];`

apps/ui/src/pages/manage/applications/index.tsx
demariadaniel: Yeah prefer we use what I've set up there, and make any changes needed to integrate

apps/ui/src/pages/manage/applications/index.tsx
surajgoraya: Yeah `as` was kind of a clutch I was using here since `TOTAL` wasn't contained anywhere in the app orgianlly as a possible state, however now it's returned from the `/metadata/counts/` API so I will move it over to use that type.

apps/ui/src/components/pages/manage/ManagementDashboard.tsx
```diff
@@ -0,0 +1,167 @@
...
+	},
+	{
+		title: 'Applicant',
+		dataIndex: ['applicantInformation', 'firstName'],
+		key: 'applicant',
+		render: (value: string, record: ApplicationWithApplicantInformation) =>
+			record.applicantInformation.firstName && record.applicantInformation.lastName
+				? `${record.applicantInformation.firstName} ${record.applicantInformation.lastName}`
+				: '-',
+		// sorter: { multiple: 3 }, NOTE: Post-MVP according to specs
```
demariadaniel, line 95: If this is unused we can remove it for now

apps/ui/src/pages/manage/applications/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+		const parsedPage = typeof pageNumber === 'string' ? parseInt(pageNumber) : pageNumber;
+
+		if (isValidPageNumber(parsedPage)) {
+			return forNetwork ? parsedPage - 1 : parsedPage;
+		}
+
+		return forNetwork ? 0 : 1;
+	} else {
+		return forNetwork ? 0 : 1;
+	}
```
demariadaniel, line 301: Returning the same thing twice, can be moved out of if/else

apps/ui/src/pages/manage/applications/index.tsx
```diff
@@ -0,0 +1,335 @@
...
+const parseFilters = (filtersFromQuery?: string | null) => {
+	if (filtersFromQuery) {
+		const decodedFilters = filtersFromQuery.split(',');
+		if (isValidFilterSet(decodedFilters)) {
+			return decodedFilters as FilterKeys[];
+		}
+		return ['TOTAL'] as FilterKeys[];
+	} else {
+		return ['TOTAL'] as FilterKeys[];
+	}
```
demariadaniel, line 320: Returning the same thing twice, and using `as` three times vs `const parseFilters = (filtersFromQuery?: string | null): FilterKeys[]` so I'd want to review how we can clear this up

apps/ui/src/pages/manage/applications/index.tsx
```diff
@@ -0,0 +1,321 @@
...
+			allUrlParams.set('page', '1');
+			setSearchParams(allUrlParams);
+			return;
+		}
+
+		/**
+		 * Even with the user having a URL-bound state, we can't true that they have a known set of filters selected.
+		 * We should validate this first before continuing.
+		 */
+		if (isFilterKeySet(parseFilters(appliedFilters)) == false) {
```
demariadaniel, line 152: This if block has the same logic as lines 142-145
So maybe we can capture the separate conditions in descriptive variables then combine the blocks into one?

```
if (missingPageState || unknownFilters) {
    allUrlParams.set('filters', 'TOTAL');
		allUrlParams.set('page', '1');
		setSearchParams(allUrlParams);
		return;
}
```

Also the comment for this block has a grammar typo, not sure what word is intended here: `we can't true that they have a known`

surajgoraya: Good catch! Will refactor that to simplify things! I had originally made this quite verbose to wrap my own head around it cause I was running into weird issues with my logic - but the code has been cleaned up (and works) now!

> Also the comment for this block has a grammar typo, not sure what word is intended here: we can't true that they have a known

Haha, I think that's supposed to say `we can't trust that they have a known set of 'good' filters selected`

apps/ui/src/api/useGetApplicationList.ts
demariadaniel: Discussed offline with @surajgoraya , we need to resolve disconnect between Drizzle Schema / Shared Types definitions to solve this properly and prevent further Type duplication. With Zod recently added we should be able to unify these.

I adding this as an action item to: https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=95531183&issue=Pan-Canadian-Genome-Library%7Cdaco%7C178
(and also including a link to this comment, so @surajgoraya leave this discussion unresolved)

demariadaniel (approved):
Looks great! Major functionality add. Tested locally, feedback addressed or planned for follow up.

PR #153, opened 2025-01-15
#59: Happy-path State transitions

demariadaniel:
## Summary

Adds ApplicationActionService DB methods to multiple StateManager functions, so that state transitions update application records and create an action record
Includes some refactoring discussed with @joneubank 

Note: State Manager Unit tests were written in such a way that Test C depends on Test B, which depends on Test A, to test that the flow worked correctly

This should be refactored but is a fair amount of work, and there are other refactors we want to cover, so we can create a follow up PR to do both
Issue #178 here: https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=95531183&issue=Pan-Canadian-Genome-Library%7Cdaco%7C178

The side effect is that multiple state transitions are updated as part of this branch, not just 'happy path':
`submitDraft, submitRepRevision, submitDacRevision, editDraft, editRepReview, editDacReview, reviseRepReview, reviseDacReview, approveRepReview, approveDacReview`
All of these will update Application + Action records
Not all state transitions are covered by this PR

### Related Issues

- [#59](https://github.com/Pan-Canadian-Genome-Library/daco/issues/59)
- [#60](https://github.com/Pan-Canadian-Genome-Library/daco/issues/60)
- [#61](https://github.com/Pan-Canadian-Genome-Library/daco/issues/61)

## Description of Changes
- Adds appropriate application/action DB methods to state transition functions
- Adds central State Manager methods for checking transition validity, and for updating DB
- Adds explicit return Types to Service functions
- Adds JsDocs to Service files
- Renames applicationApi to applicationController
- Migration to add 'Edit' Action type

### Special Instructions

Schema updates require running `pnpm drizzle-kit migrate`

`api` folder has been changed to `controllers`, `applicationApi` changed to `applicationController`

## Readiness Checklist

- [x] **Self Review**
  - I have performed a self review of code
  - I have run the application locally and manually tested the feature
- [x] **PR Format**
  - The PR title is properly formatted to match the pattern: `#{TicketNumber}: Description of Changes`
  - Links are included to all relevant tickets
- [x] **Labels Added**
  - Label is added for each package/app that is modified (`api`, `ui`, `data-model`, etc.)
  - Label is added for the type of work done in this PR (`feature`, `fix`, `chore`, `documentation`)
- [x] **Local Testing**
  - Successfully built all packages locally
  - Successfully ran all test suites, all unit and integration tests pass
- [x] **Updated Tests**
[...]

PR #175, opened 2025-01-27
WIP: Feat/167 add collaborators endpoint

demariadaniel:
### Related Issues

- Paste URL to issue 1
- Paste URL to issue 2

## Readiness Checklist

- [ ] **Self Review**
- [ ] **PR Format**
- [ ] **Labels Added**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation

### DIFF
--- a/apps/api/src/routes/application-router.ts
+++ b/apps/api/src/routes/application-router.ts
@@ -33,36 +33,6 @@ import { isPositiveNumber } from '@/utils/routes.js';
 const applicationRouter = express.Router();
 const jsonParser = bodyParser.json();
 
-/**
- * TODO:
- * 	- Currently no validation is done to ensure that the current logged in user can create a application. This should be done and refactored.
- * 	- Validate request params using Zod.
- */
-applicationRouter.post(
-	'/applications/create',
-	jsonParser,
-	async (request: Request<{}, {}, { userId: string }, any>, response) => {
-		const { userId } = request.body;
-
-		/**
-		 * TODO: Temporary userId check until validation/dto flow is confirmed.
-		 * Reflect changes in swagger once refactored.
-		 **/
-		if (!userId) {
-			response.status(400).send({ message: 'User ID is required.' });
-			return;
-		}
-
-		const result = await createApplication({ user_id: userId });
-
-		if (result.success) {
-			response.status(201).send(result.data);
-		} else {
-			response.status(500).send({ message: result.message, errors: String(result.errors) });
-		}
-	},
-);
-
 applicationRouter.post('/applications/edit', jsonParser, async (req, res) => {
 	// TODO: Add Auth & Zod validation
 	const data = req.body;
@@ -188,6 +158,7 @@ applicationRouter.get('/applications/metadata/counts', async (req: Request<{}, {
 		res.status(500).send({ message: result.message, errors: String(result.errors) });
 	}
 });
+
 applicationRouter.post('/applications/approve', jsonParser, async (req, res) => {
 	const { applicationId }: { applicationId?: number } = req.body;
 
@@ -232,5 +203,59 @@ applicationRouter.post('/applications/approve', jsonParser, async (req, res) => 
 		});
 	}
 });
+
+/**
+ * TODO:
+ * 	- Currently no validation is done to ensure that the current logged in user can create a application. This should be done and refactored.
+ * 	- Validate request params using Zod.
+ */
+applicationRouter.post(
+	'/applications/create',
+	jsonParser,
+	async (request: Request<{}, {}, { userId: string }, any>, response) => {
+		const { userId } = request.body;
+
+		/**
+		 * TODO: Temporary userId check until validation/dto flow is confirmed.
+		 * Reflect changes in swagger once refactored.
+		 **/
+		if (!userId) {
+			response.status(400).send({ message: 'User ID is required.' });
+			return;
+		}
+
+		const result = await createApplication({ user_id: userId });
+
+		if (result.success) {
+			response.status(201).send(result.data);
+		} else {
+			response.status(500).send({ message: result.message, errors: String(result.errors) });
+		}
+	},
+);
+
+/**
+ * Add Collaborator
+ */
+applicationRouter.post(
+	'/applications/collaborators',
+	jsonParser,
+	async (request: Request<{}, {}, { userId: string }, any>, response) => {
+		const { userId } = request.body;
+
+		if (!userId) {
+			response.status(400).send({ message: 'User ID is required.' });
+			return;
+		}
+
+		const result = await createApplication({ user_id: userId });
+
+		if (result.success) {
+			response.status(201).send(result.data);
+		} else {
+			response.status(500).send({ message: result.message, errors: String(result.errors) });
+		}
+	},
+);
 
 export default applicationRouter;
--- a/apps/api/src/routes/application-router.ts
+++ b/apps/api/src/routes/application-router.ts
@@ -234,28 +234,4 @@ applicationRouter.post(
 	},
 );
 
-/**
- * Add Collaborator
- */
-applicationRouter.post(
-	'/applications/collaborators',
-	jsonParser,
-	async (request: Request<{}, {}, { userId: string }, any>, response) => {
-		const { userId } = request.body;
-
-		if (!userId) {
-			response.status(400).send({ message: 'User ID is required.' });
-			return;
-		}
-
-		const result = await createApplication({ user_id: userId });
-
-		if (result.success) {
-			response.status(201).send(result.data);
-		} else {
-			response.status(500).send({ message: result.message, errors: String(result.errors) });
-		}
-	},
-);
-
 export default applicationRouter;
--- a/apps/api/src/routes/collaborators-router.ts
+++ b/apps/api/src/routes/collaborators-router.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 The Ontario Institute for Cancer Research. All rights reserved
+ * Copyright (c) 2025 The Ontario Institute for Cancer Research. All rights reserved
  *
  * This program and the accompanying materials are made available under the terms of
  * the GNU Affero General Public License v3.0. You should have received a copy of the
@@ -17,27 +17,36 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { relations } from 'drizzle-orm';
-import { bigint, pgTable, text, varchar } from 'drizzle-orm/pg-core';
-import { applicationContents } from './applicationContents.js';
+import bodyParser from 'body-parser';
+import express, { Request } from 'express';
 
-export const collaborators = pgTable('collaborators', {
-	id: bigint({ mode: 'number' }).primaryKey().generatedAlwaysAsIdentity(),
-	application_id: bigint({ mode: 'number' }),
-	first_name: varchar({ length: 255 }).notNull(),
-	middle_name: varchar({ length: 255 }),
-	last_name: varchar({ length: 255 }).notNull(),
-	title: varchar({ length: 255 }),
-	suffix: varchar({ length: 255 }),
-	position_title: varchar({ length: 255 }).notNull(),
-	institutional_email: varchar({ length: 320 }).notNull(),
-	profile_url: text(),
-	collaborator_type: text(),
-});
+import { createApplication } from '@/api/application-api.js';
 
-export const collaboratorsRelations = relations(collaborators, ({ one }) => ({
-	application_id: one(applicationContents, {
-		fields: [collaborators.application_id],
-		references: [applicationContents.application_id],
-	}),
-}));
+const collaboratorsRouter = express.Router();
+const jsonParser = bodyParser.json();
+
+/**
+ * Add Collaborator
+ */
+collaboratorsRouter.post(
+	'/collaborators',
+	jsonParser,
+	async (request: Request<{}, {}, { userId: string }, any>, response) => {
+		const { userId } = request.body;
+
+		if (!userId) {
+			response.status(400).send({ message: 'User ID is required.' });
+			return;
+		}
+
+		const result = await createApplication({ user_id: userId });
+
+		if (result.success) {
+			response.status(201).send(result.data);
+		} else {
+			response.status(500).send({ message: result.message, errors: String(result.errors) });
+		}
+	},
+);
+
+export default collaboratorsRouter;
--- a/apps/api/src/routes/collaborators-router.ts
+++ b/apps/api/src/routes/collaborators-router.ts
@@ -20,7 +20,7 @@
 import bodyParser from 'body-parser';
 import express, { Request } from 'express';
 
-import { createApplication } from '@/api/application-api.js';
+import { createCollaborators } from '@/api/collaborators-api.js';
 
 const collaboratorsRouter = express.Router();
 const jsonParser = bodyParser.json();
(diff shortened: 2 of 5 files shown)


apps/api/src/routes/application-router.ts
```diff
@@ -33,36 +33,6 @@ import { isPositiveNumber } from '@/utils/routes.js';
 const applicationRouter = express.Router();
 const jsonParser = bodyParser.json();
 
-/**
```
demariadaniel, line 36: Moved for alphabetization

apps/ui/src/global/types.ts
```diff
@@ -29,16 +29,6 @@ export interface Application {
 	updatedAt: string;
 	expiresAt: string;
 }
-export interface ApplicationWithApplicantInformation extends Application {
```
demariadaniel, line 32: These changes (and related changes in `data-model` / `manage/applications`) are also in :https://github.com/Pan-Canadian-Genome-Library/daco/pull/153
Intent is to make types that are shared and reusable btwn ui/api

apps/api/package.json
```diff
@@ -30,11 +30,11 @@
 		"@types/body-parser": "^1.19.5",
 		"body-parser": "^1.20.3",
 		"cors": "^2.8.5",
-		"drizzle-orm": "^0.38.3",
+		"drizzle-orm": "^0.39.1",
```
demariadaniel, line 33: See note in `service/collaboratorsService` https://github.com/Pan-Canadian-Genome-Library/daco/pull/175/files#diff-fe99e87f7017cc3474b7915429480c6fb9eb76d78603256f7645fa59ce2a2d6e

I updated this to attempt fixing a bug (bug persists), could be removed if there's any issue

apps/api/tests/api/collaborators-controller.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+			const result = await createCollaborators({ application_id, user_id, collaborators });
+
+			assert.ok(result.success && result.data[0]);
+
+			const newCollaboratorRecord = result.data[0];
+
+			assert.strictEqual(newCollaboratorRecord.application_id, application_id);
+		});
+	});
```
demariadaniel, line 77: Intend to add tests for validation / failure cases before merging

apps/api/src/api/collaboratorsController.ts
```diff
@@ -0,0 +1,100 @@
...
+
+	// TODO: Add Real Auth
+	// Validate User is Applicant
+	if (!(user_id === application.user_id)) {
+		return failure('Unauthorized, cannot create Collaborators');
+	}
+
+	if (!(application.state === 'DRAFT')) {
+		return failure(`Can only add Collaborators when Application is in state DRAFT`);
+	}
```
demariadaniel, line 71: There is no state transition involved here but we may want to leverage `state manager` here when it's merged

demariadaniel:
Screenshot 2025-02-05 at 6 03 19 PM

Added first tests for validations (incorrect user id, incorrect application id, missing fields)

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -0,0 +1,100 @@
...
+}: {
+	application_id: number;
+	user_id: string;
+	collaborators: CollaboratorDTO[];
+}) => {
+	const database = getDbInstance();
+	const collaboratorsRepo: CollaboratorsService = collaboratorsSvc(database);
+	const applicationRepo: ApplicationService = applicationSvc(database);
+
+	const applicationResult = await applicationRepo.getApplicationById({ id: application_id });
```
joneubank, line 55: Before fetching the app we should perform our checks if this function should run, ie. do the auth check first. (I know auth is a TODO atm but lets put it in the right place).

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -0,0 +1,100 @@
...
+		return failure(`Can only add Collaborators when Application is in state DRAFT`);
+	}
+
+	const validCollaborators: ValidCollaboratorDTO[] = collaborators.filter(
+		(data: CollaboratorDTO): data is ValidCollaboratorDTO =>
+			!!data.collaboratorFirstName &&
+			!!data.collaboratorLastName &&
+			!!data.collaboratorPositionTitle &&
+			!!data.collaboratorInstitutionalEmail,
+	);
```
joneubank, line 79: Use a zod schema.

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -0,0 +1,100 @@
...
+		last_name: data.collaboratorLastName,
+		suffix: data.collaboratorSuffix,
+		position_title: data.collaboratorPositionTitle,
+		institutional_email: data.collaboratorInstitutionalEmail,
+		application_id,
+	}));
+
+	const result = await collaboratorsRepo.createCollaborators({
+		newCollaborators,
+	});
```
joneubank, line 97: Do we need to do any checks for duplicate collaborators?

joneubank (changes requested):

apps/api/src/controllers/collaboratorsController.ts
demariadaniel: Wondered the same thing

apps/api/src/controllers/collaboratorsController.ts
demariadaniel: This is confirming that the application id and user id match, and you can only do that by comparing what's in the request body with what's in the database

So I've built it to retrieve a record, confirm the request details match; if they don't match you error here; if they do match you proceed with the rest of the application actions

This was a solution to putting in 'working' auth until we have a different approach defined
So I could move this check but it won't actually be checking the request body against 'real' data

apps/api/src/controllers/collaboratorsController.ts
demariadaniel: Thanks! Updated w/ related changes

apps/api/src/routes/collaborators-router.ts
```diff
@@ -0,0 +1,75 @@
...
+const jsonParser = bodyParser.json();
+
+/**
+ * Add Collaborator
+ */
+collaboratorsRouter.post(
+	'/collaborators/create',
+	jsonParser,
+	async (request: Request<{}, {}, CollaboratorRequest, any>, response) => {
+		const { applicationId: application_id, userId: user_id, collaborators } = request.body;
```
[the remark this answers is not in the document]
demariadaniel: OK I've added req body Zod schema validation, which ends up moving the validation for the `collaborators` array into the router.
Unit testing for this validation will have to be revisited when we add mock requests for the tests.

apps/api/src/controllers/collaboratorsController.ts
demariadaniel: So -- there's one layer of auth validation on the router (is userId included). 
Then this line in the controller adds a second layer of validation with the db (userId and applicationId are a match). Both are to be replaced when real Auth is added.

Let me know whether to keep both, or reorganize, or what you're thinking

apps/api/src/controllers/collaboratorsController.ts
demariadaniel: Added handling for preventing duplicates in req.body, now adding db layer

demariadaniel: OK I've added logic to prevent creating duplicates in the database
Required quite a few changes which is why I didn't finish last night:
[Prevenet Duplicate Collaborator Records](https://github.com/Pan-Canadian-Genome-Library/daco/pull/175/commits/0b58ea4a76efa22ccdc441d2b8ea6ef1d68520c3)

demariadaniel:
Tests passing post-PR feedback updates
Screenshot 2025-02-13 at 11 57 36 AM

apps/api/src/service/collaboratorsService.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { type PostgresDb } from '@/db/index.js';
+import { collaborators } from '@/db/schemas/collaborators.js';
+import logger from '@/logger.js';
+import { failure, success } from '@/utils/results.js';
+import { and, eq } from 'drizzle-orm';
+import { type CollaboratorModel, type CollaboratorRecord } from './types.js';
+
+const collaboratorsSvc = (db: PostgresDb) => ({
+	createCollaborators: async ({ newCollaborators }: { newCollaborators: CollaboratorModel[] }) => {
```
demariadaniel, line 28: Need to add explicit return type

pnpm-lock.yaml
demariadaniel, line 1: Lockfile updates to fix Drizzle conflicts: https://github.com/drizzle-team/drizzle-orm/issues/2418

apps/api/src/routes/collaboratorsRouter.ts
```diff
@@ -0,0 +1,87 @@
...
+			}
+
+			if (errorField === 'userId') {
+				// TODO: Add Real Auth
+				response.status(401).send({ message: 'Unauthorized, cannot create Collaborators' });
+				return;
+			}
+
+			if (errorField === 'applicationId') {
+				response.status(404).send({ message: 'applicationId is missing, cannot create Collaborators' });
```
JamesTLopez, line 80: If I put a string "4" instead of a number it will return this error, which is not entirely correct and might be misleading. Maybe make the errors here more generic or type check before it gets here?

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -0,0 +1,97 @@
...
+	if (!(user_id === application.user_id)) {
+		return failure('Unauthorized, cannot create Collaborators', 'Unauthorized');
+	}
+
+	if (!(application.state === 'DRAFT')) {
+		return failure(`Can only add Collaborators when Application is in state DRAFT`, 'InvalidState');
+	}
+
+	const hasDuplicateRecords = collaborators.some((collaborator, index) => {
+		const matchingRecordIndex = collaborators.findIndex(
```
joneubank, line 66: Nested loops is a code flag for performance. Would want to check performance if there were 100s of collaborators.

apps/api/src/service/collaboratorsService.ts
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright (c) 2024 The Ontario Institute for Cancer Research. All rights reserved
```
JamesTLopez, line 2: 2025

apps/api/src/routes/collaboratorsRouter.ts
```diff
@@ -0,0 +1,87 @@
...
+const jsonParser = bodyParser.json();
+
+/**
+ * Add Collaborator
+ */
+collaboratorsRouter.post(
+	'/collaborators/create',
+	jsonParser,
+	async (request: Request<{}, {}, CollaboratorRequest, any>, response) => {
+		const validatedPayload = collaboratorsRequestSchema.safeParse(request.body);
```
joneubank, line 37: Can we use the request middleware package?
- https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/packages/request-utils/src/validation.ts#L53

Example usage: https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/apps/api/src/routes/application-router.ts#L73

apps/api/src/service/collaboratorsService.ts
```diff
@@ -0,0 +1,92 @@
...
+import { type AsyncResult, failure, success } from '@/utils/results.js';
+import { and, eq } from 'drizzle-orm';
+import { type CollaboratorModel, type CollaboratorRecord } from './types.js';
+
+const collaboratorsSvc = (db: PostgresDb) => ({
+	createCollaborators: async ({
+		newCollaborators,
+	}: {
+		newCollaborators: CollaboratorModel[];
+	}): AsyncResult<CollaboratorRecord[]> => {
```
joneubank, line 32: Should this accept transaction as parameter?

apps/api/src/service/collaboratorsService.ts
```diff
@@ -0,0 +1,92 @@
...
+
+					newRecords.push(newCollaboratorRecord[0]);
+				});
+
+				return newRecords;
+			});
+
+			if (!(collaboratorRecords.length === newCollaborators.length)) {
+				throw new Error(`Error creating new collaborators: ${collaboratorRecords}`);
+			}
```
joneubank, line 78: I'll update this with my Failure type changes in my auth PR... in the future, this is the purpose of the failure type and we can return that instead of throwing error. Failures can communicate expected error situations instead of escaping out into undocumented thrown exceptions.

apps/api/tests/api/collaborators-controller.test.ts
joneubank, line 1: Good test cases.

packages/data-model/src/types.ts
joneubank, line 1: We need to stop having generic "types" files.

I don't want to change it in this PR but from now on lets organize our types into descriptively named files, or export types from the files where they are created and used.

joneubank (dismissed):
I think this is functional, thanks for the test cases. I think we can iterate on some of the code organization but I want to first get this functionality merged.

apps/api/src/routes/collaboratorsRouter.ts
```diff
@@ -0,0 +1,87 @@
...
+				response.send({ message, errors });
+				return;
+			}
+		} else {
+			const { issues } = validatedPayload.error;
+			const errorField = issues[0]?.path[0];
+
+			if (errorField === 'collaborators') {
+				response.status(400).send({ message: 'Required Collaborator details are missing.' });
+			}
```
JamesTLopez, line 71: If I trigger the zod validation here for `tooFew` which is putting in only 1 character, it will display error
"Required Collaborator details are missing". 

I think we should make this more generic to be more clear. Since this is a zod validation related error, maybe we can use something like "A collaborator(s) requirements do not fit the accepted format"

demariadaniel: I was also thinking adding some of the Zod error info (i.e. adding `const errorMessage = issues[0]?.message;` & including it in the message) could be useful here

JamesTLopez: > I was also thinking adding some of the Zod error info (i.e. adding `const errorMessage = issues[0]?.message;` & including it in the message) could be useful here

Even better!

apps/api/src/routes/collaboratorsRouter.ts
demariadaniel: Added, major improvement

apps/api/src/routes/collaboratorsRouter.ts
demariadaniel: I've added `withSchemaValidation` which provides better error messaging and type checking
Screenshot 2025-02-27 at 2 52 07 PM

apps/api/src/routes/collaboratorsRouter.ts
demariadaniel: Handled by use of `withSchemaValidation`

apps/api/src/service/collaboratorsService.ts
demariadaniel: The thinking here is that DB actions can (and do) throw errors so it's necessary to wrap the whole block in a `try/catch`
So I funneled all failures into the `catch` block which always returns a `failure`, with error logging

apps/api/src/service/collaboratorsService.ts
demariadaniel: Potentially, there's currently no use case for creating Collaborators simultaneously with another DB action

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -0,0 +1,97 @@
...
+			(record, searchIndex) =>
+				searchIndex !== index &&
+				record.collaboratorFirstName === collaborator.collaboratorFirstName &&
+				record.collaboratorLastName === collaborator.collaboratorLastName &&
+				record.collaboratorInstitutionalEmail === collaborator.collaboratorInstitutionalEmail &&
+				record.collaboratorPositionTitle === collaborator.collaboratorPositionTitle,
+		);
+
+		return matchingRecordIndex !== -1;
```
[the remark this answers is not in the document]
demariadaniel: Updated

apps/api/src/controllers/collaboratorsController.ts
demariadaniel: We will discuss offline - this is a fair call out but you are comparing objects (so `objectA === objectB` won't work) and hypothetically different collaborators could have identical first names, last names, or position titles

I'm looking at the following as a potential solution:
```
const collaboratorEmails = new Set(collaborators.map((record) => record.collaboratorInstitutionalEmail));
const hasDuplicateRecords = collaboratorEmails.size !== collaborators.length;
```

JamesTLopez (approved):

Issue #161, opened 2025-01-22
Form Validation - Applicant Information page


Buwujiu:
## Descriptions
Implement form validation for Applicant Information page. The validation should ensure the following rules:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-58263b74-7fff-04a7-56ad-ab8037ab1eaf"><ul style="margin-top:0;margin-bottom:0;padding-inline-start:48px;"><li dir="ltr" style="list-style-type:disc;font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;margin-left: 36pt;" aria-level="1"></li></ul><div dir="ltr" style="margin-left:36pt;" align="left">
Title | Dropdown Values: Dr., Miss, Mr., Mrs., Ms., Prof.
-- | --
All Required fields | Indicated with red *. Must be filled in, if left empty, and when cursor pointing away, should immediately display an error message “Please fill out the required field.” in red.
Primary Affiliation | Free text field
Institutional Email | UI Should perform a basic format validation on email address. Enruses: Presence of an @ symbol.A valid local part (before @).A domain part (after @) with a top-level domain (e.g., .com, .org).if invalid, display error message “Please enter a valid email address.” in red.
Researcher Profile | UI Should perform basic format validation. Ensures that the URL follows a standard format, such as http://example.com or https://example.com.Requirements:Starts with a scheme (e.g., http, https).Contains a valid domain or IP address.May include optional ports, paths, queries, or fragments.If invalid, display an error message “Please enter a valid url. Must begin with http:// or https://.” in red.
Position Title | Free text field
Country | Realistically we will only need Canada, but in case PCGL opens up for international projects, we should support a full country list. Display a dropdown menu of alphabetically ordered country list, sorted by English or French name, depending on the selected site language.Make “Canada” the first default option on top of the list. Nice to have: make the list searchable.
Street Address | Free text field
City and Province | Free text field
Building | Free text field
Postal/Zip code | Free text field

</div></b>

Refer to https://docs.google.com/document/d/1NK5j-ZgwPDwkUmMvC3KfNIdXAhm8JrWmSWBekOtN_SM/edit?tab=t.0#heading=h.5dd7jtld668

PR #177, opened 2025-01-28
#128: Applicant Information Validation and PCGL Validation Package

JamesTLopez:
## Summary

Create the Zod Schemas for the validation on relevant fields on the Applicant Information page.
Create Validation pcgl package to house validation rules between backend and frontend.

### Related Issues

- #161

## Description of Changes

- create validation rules for applicant information
- create custom pcgl validation package 
- create CustomFormErrorTranslationMapping
   -  intercepts errors allowing use to customize our error messages rendered on the antd components
- create enFormErrors and frFormErrors i18n translations

## Translating Zod Errors on the UI

A challenge encountered here is the translations of zod errors on the UI side. OHCRN solution was not entirely applicable to our application, they have a custom i18n solution and did not entirely rely on a CSS framework to handle the UI components. 

Out of the box, Antd has error UI built into the form components which can be triggered by applying the validation [rules](https://ant.design/components/form#rule). The error component inside antd form elements cannot be customized (only using themes). We cannot directly translate the text in the component like what is done on OHCRN (like `<Error/>` ). 

The current solution proposed in the PR is to intercept the incoming errors from zod using `z.setErrorMap(...);` and applying translations through the custom error mapping. That way we can keep our validation package errors generic, and only apply the custom translations required only on the UI side which also defaults back into the generic Zod Errors if not otherwise specified. 

The validation package will trigger the same validation between the Api and UI, but the UI can dictate the messages it displays without compromising the backend

```js
const CustomFormErrorTranslationMapping: z.ZodErrorMap = (error, ctx) => {
	/**
	 * Add translations through here
	 */
	switch (error.code) {
		case z.ZodIssueCode.invalid_type:
			if (error.expected === 'string') {
				return { message: i18n.t('requiredField') };
			}
			break;
		case z.ZodIssueCode.invalid_string:
			if (error.validation === 'email') {
				return { message: i18n.t('validEmail') };
			} else if (error.validation === 'url') {
				return { message: i18n.t('validURL') };
			} else if (error.validation === 'regex') {
				// right now this only does the postal code, when we expand 'regex' outside of postal code, we need to adjust here
				return { message: i18n.t('validPostalCode') };
			}
			break;
	}
	return { message: ctx.defaultError };
};
```

### Special Instructions
You will need to build `@pcgl-daco/validation`, so go to the top level folder `daco/*` and run:
```
pnpm build:all
```

and you will need to install `@pcgl-daco/validation` on the UI repo:
```
pnpm i
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
  - 
image

- [x] **Updated Tests**
- [x] Documentation


packages/pcgl-validation/src/validation.ts
```diff
@@ -0,0 +1,65 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import * as z from 'zod';
+
+export type ApplicantInformationType = {
```
surajgoraya, line 22: Not a major thing, but would it a bit neater to split up the types and validation schemas into two separate TS files? I think it makes things a little more readable as this package will likely grow.

packages/pcgl-validation/src/validation.ts
```diff
@@ -0,0 +1,47 @@
...
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import * as z from 'zod';
+
+export const applicantInformationSchema = z.object({
+	applicantTitle: z.string({ message: 'Please fill out the required field' }),
+	applicantFirstName: z
+		.string({ message: 'Please fill out the required field' })
```
surajgoraya, line 25: Not a change, but I'm curious how we should handle i18n with splitting the validaiton schemas into a sperate package?

JamesTLopez: Oh good point, I didn't think of that! Lets look into that

apps/ui/src/components/pages/application/SectionFooter.tsx
```diff
@@ -56,6 +56,10 @@ const SectionFooter = ({ currentRoute, isEditMode }: SectionFooterProps) => {
 	};
 
 	const nextSection = () => {
+		if (!!onSubmit) {
+			onSubmit();
```
Azher2Ali, line 60: Can we consider adding default behavior when onSubmit is not provided (e.g., a simple validation check or fallback).

packages/pcgl-validation/src/validation.ts
```diff
@@ -0,0 +1,47 @@
...
+		.url({ message: 'Please enter a valid Url address.' })
+		.refine((val) => val.startsWith('https://') || val.startsWith('http://'), {
+			message: 'Please enter a valid url. Must begin with http:// or https://.',
+			path: ['url'],
+		}),
+	applicantPositionTitle: z.string({ message: 'Please fill out the required field' }),
+	institutionCountry: z.string({ message: 'Please fill out the required field' }),
+	institutionState: z.string({ message: 'Please fill out the required field' }),
+	institutionCity: z.string({ message: 'Please fill out the required field' }),
+	institutionPostalCode: z.string({ message: 'Please fill out the required field' }),
```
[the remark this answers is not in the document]
JamesTLopez: I think we decided in a project planning meeting that the postal code would be a free text field since they can be very different depending on the country they are applying from. Also specified in the [DACO Functional specs](https://docs.google.com/document/d/1NK5j-ZgwPDwkUmMvC3KfNIdXAhm8JrWmSWBekOtN_SM/edit?tab=t.0)

@demariadaniel @surajgoraya Am I wrong about this?

apps/ui/src/components/pages/application/SectionFooter.tsx
JamesTLopez: It should default back to just going to the next page if onSubmit was not passed down from the parent component:

```js

const nextSection = () => {
if (!!onSubmit) {
	onSubmit();
	return;
}
navigate(`/application/${id}/${nextRoute}/${isEditMode ? 'edit' : ''}`, { replace: true });
};

```

To be honest, this is probably going to get a refactor once we start adding the `application store` and the api calls to update the application on the backend. This is purely just to trigger the validation rules for ant-d.

Azher2Ali: That sounds great! I agree with the approach of defaulting to navigating to the next page when onSubmit is not provided. It keeps things simple for now, especially while we’re triggering the validation rules. Once we integrate the application store and backend API calls, we can refactor this for more comprehensive handling. Thanks for the input!

packages/pcgl-validation/src/validation.ts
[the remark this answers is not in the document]
surajgoraya: There does exist a postal code validation regex for Canadian Postal codes see OHCRN's postal code regex here: https://github.com/OHCRN/platform/blob/develop/packages/types/src/common/regexes.ts#L33, however, I'm wondering if we should apply a regex at all? Is it possible we get non-Canadian insitutions? I think while Canada and the UK have similar postal codes, the US' are completely different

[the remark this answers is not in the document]
JamesTLopez: > however, I'm wondering if we should apply a regex at all?

Just to hop in on this, I think for now, we should just restrict the user from using special characters for now till specified otherwise. Pushing this change

packages/pcgl-validation/src/validation.ts
JamesTLopez: Implemented the solution, refer to PR outline for explanation

apps/ui/src/i18n/translations.ts
```diff
@@ -19,26 +19,65 @@
...
+ * Translation error mapping for custom error messages
+ *  NOTE: Question translating the other default errors options:
+ *  - 1. we install the zod-translation package
+ *  - 2. we add translations as we go from the CustomFormErrorTranslationMapping
+ */
+const CustomFormErrorTranslationMapping: z.ZodErrorMap = (error, ctx) => {
+	/**
+	 * Add translations through here
+	 */
+	switch (error.code) {
```
surajgoraya, line 61: As discussed yesterday, it might be worth noting in a comment or something that if a user provides a `message` attribute to the zod error, it'll break this translation function?

surajgoraya (approved):
These look good to me! The way we're doing translation is pretty easy to follow given our current constraints with `antd` and how our project is structured.

packages/pcgl-validation/src/schemas.ts
```diff
@@ -0,0 +1,42 @@
...
+// Applicant Information Form Section
+export type ApplicantInformationSchemaType = z.infer<typeof applicantInformationSchema>;
+export const applicantInformationSchema = z.object({
+	applicantTitle: NonEmptyString,
+	applicantFirstName: NonEmptyString,
+	applicantMiddleName: EmptyOrOptionalString,
+	applicantLastName: NonEmptyString,
+	applicantSuffix: EmptyOrOptionalString,
+	applicantPrimaryAffiliation: NonEmptyString,
+	applicantInstituteAffiliation: NonEmptyString.email(),
```
demariadaniel, line 33: I don't think `applicantInstituteAffiliation` is the right key name?

apps/ui/src/pages/applications/sections/applicant.tsx
```diff
@@ -202,7 +176,7 @@ const Applicant = () => {
 						<Col xs={{ flex: '100%' }} md={{ flex: '100%' }} lg={{ flex: '50%' }}>
 							<InputBox
 								label={translate('applicant-section.form.country')}
-								name="institutionCountry"
+								name="applicantInstituteCountry"
```
JamesTLopez, line 179: Will turn this into a select box, Suraj has a working solution, will refactor after his PR is merged in.

apps/ui/src/pages/applications/sections/applicant.tsx
```diff
@@ -18,12 +18,12 @@
  */
 
 import { zodResolver } from '@hookform/resolvers/zod';
+import { applicantInformationSchema, ApplicantInformationSchemaType } from '@pcgl-daco/validation';
```
demariadaniel, line 21: Can you add `type ApplicantInformationSchemaType` here? Jon and I were chatting yesterday, it's a helpful addition

packages/pcgl-validation/src/schemas.ts
```diff
@@ -0,0 +1,42 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { z } from 'zod';
+import { EmptyOrOptionalString, NonEmptyString } from './common/strings.js';
```
demariadaniel, line 21: Also here `import { type EmptyOrOptionalString, type NonEmptyString }`

packages/pcgl-validation/package.json
demariadaniel, line 1: We could change the folder name to just `validation`, it's all part of the `pcgl-daco` project so I don't think there's any distinction btwn this and any other kind of validation

packages/pcgl-validation/src/schemas.ts
JamesTLopez: I'll need those as zod fields, not as types to create the proper fields in the schema. Unless i'm missing something here.

demariadaniel: I see, my bad

demariadaniel: Got mixed up with the duplicate type/const names

JamesTLopez: All good! 😎

packages/pcgl-validation/package.json
JamesTLopez: This makes sense, applying changes

apps/ui/src/pages/applications/sections/applicant.tsx
```diff
@@ -162,7 +136,7 @@ const Applicant = () => {
 							<InputBox
 								label={translate('applicant-section.form.primaryEmail')}
 								subLabel={translate('applicant-section.form.primaryEmailLabel')}
-								name="applicantInstituteAffiliation"
+								name="applicantInstituteEmail"
 								control={control}
```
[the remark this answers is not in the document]
demariadaniel: OK -- it's resolved, may have run `build` before pulling latest

demariadaniel (approved):
OK it's looking great to have this working and set up to use!

Issue #119, opened 2024-12-18
Endpoint - DAC Reject Application


Buwujiu:
## Descriptions
Build a REST endpoint to allow DAC reject application. 
Rejected applications cannot be re-opened or changed.

Buwujiu:
Code is done, for testing, waiting for the submit App endpoint to be merged.

Buwujiu:
Will test in dev using UI. Closing ticket!

PR #179, opened 2025-01-29
#119: Add REST endpoint for DAC to reject applications 

Azher2Ali:
## Summary

This PR introduces a new REST API endpoint to allow DAC  to reject applications. Once an application is rejected, it cannot be reopened or modified. The application status transitions from DAC Review to Rejected, and a timestamp is recorded in the application history. The API returns a clear success or failure message. (#119)

## Description of Changes

**API Service**

- Added a new POST /applications/{applicationId}/reject endpoint for DAC to reject applications.

- Ensured that DAC can only reject one application at a time.

- Updated the application status transition logic:

     1. DAC Review → Rejected
     2. Applications that are Rejected cannot be reopened or modified.

- Implemented history tracking by appending a timestamp when an application is rejected.

- Standardized API response messages for success and failure scenarios.

**Validation & Constraints**

- Ensured that only applications in DAC Review status can be rejected.
- Applications already in Rejected status cannot be modified further.
- Restricted the action to authorized DAC users only.

- Database Changes

Added a timestamp field to track rejection events in the application history table.

## Readiness Checklist

- [x] **Self Review**
  - I have reviewed the code and ensured it meets the requirements.
  - The feature has been manually tested locally.
- [x] **PR Format**
  - The PR title is properly formatted to match the pattern: `#{119}: feature: Add REST endpoint for DAC to reject applications`
- [x] **Labels Added**
  - `api, feature` labels have been applied.
- [x] **Local Testing**
  
tests


apps/api/src/routes/application-router.ts
```diff
@@ -233,4 +235,50 @@ applicationRouter.post('/applications/approve', jsonParser, async (req, res) =>
...
+	const { applicationId } = req.body;
+
+	if (!applicationId) {
+		res.status(400).json({ message: 'Application ID is required.' });
+	}
+
+	if (typeof applicationId !== 'number' || !applicationId) {
+		res.status(400).send({
+			message: 'Invalid request. ApplicationId must be a valid number and is required.',
+			errors: 'MissingOrInvalidParameters',
```
surajgoraya, line 248: This is likely out of the scope of this PR, so feel free to ignore this, but I wonder if we should add a `code` object into this return which is specifically for a error code?

We also likely should make these a centralized thing, but  again, that's likely out of the scope of this PR

apps/api/src/routes/application-router.ts
```diff
@@ -233,4 +235,50 @@ applicationRouter.post('/applications/approve', jsonParser, async (req, res) =>
...
+	if (result.success) {
+		const { state } = result.data;
+
+		if (state === ApplicationStates.REJECTED) {
+			res.status(400).json({ message: 'Application is already rejected and cannot be modified.' });
+		}
+
+		if (state !== ApplicationStates.DAC_REVIEW) {
+			res.status(400).json({ message: 'Application must be in DAC Review status to be rejected.' });
+		}
```
surajgoraya, line 264: These two if blocks basically do the same thing, I wonder if it's worth it to specifically do a check for a application is already rejected? We could just provide the !== DAC_REVIEW message since it covers all scenarios.

apps/api/src/resources/swagger.yaml
```diff
@@ -317,7 +317,59 @@ paths:
...
+                        format: date-time
+        400:
+          description: Invalid request or rejection not allowed
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
```
surajgoraya, line 361: You likely should just reference the `ClientErrors` scheme here? See other API defs in the file for a example.
```
content:
   application/json:
        schema:
              $ref: '#/components/responses/ClientErrors'
```

surajgoraya (commented):
Overall it looks solid! Great work :) 

Just a few small suggestions + discussion points

apps/api/src/resources/swagger.yaml
```diff
@@ -317,7 +317,59 @@ paths:
...
+          description: Server error
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                  details:
+                    type: string
```
surajgoraya, line 372: Sorry, similar thing here, it'll be a good idea to just reference the `SeverError` schema here too. Might be a good idea to just copy and paste from a previously defined API, this helps standardize things :)

apps/api/src/routes/application-router.ts
```diff
@@ -233,4 +235,50 @@ applicationRouter.post('/applications/approve', jsonParser, async (req, res) =>
...
+		const rejectedApplication = await rejectApplication({ applicationId });
+
+		if (rejectedApplication.success) {
+			res.status(200).send({
+				message: 'Application rejected successfully.',
+				data: rejectedApplication.data,
+			});
+		} else {
+			let message = rejectedApplication.message || 'An unexpected error occurred.';
+			let errors = rejectedApplication.errors;
```
JamesTLopez, line 276: You can probably use `const` here.

apps/api/src/routes/application-router.ts
JamesTLopez: > These two if blocks basically do the same thing, I wonder if it's worth it to specifically do a check for a application is already rejected? We could just provide the !== DAC_REVIEW message since it covers all scenarios.

Should we make the message generic in this case?

apps/api/src/routes/application-router.ts
```diff
@@ -233,4 +235,50 @@ applicationRouter.post('/applications/approve', jsonParser, async (req, res) =>
...
 });
 
+applicationRouter.post('/applications/reject', jsonParser, async (req, res) => {
+	const { applicationId } = req.body;
+
+	if (!applicationId) {
+		res.status(400).json({ message: 'Application ID is required.' });
+	}
+
+	if (typeof applicationId !== 'number' || !applicationId) {
```
JamesTLopez, line 245: You can probably simplify this condition using something like this:

```js
if (isNaN(parseInt(applicationId))) {
...
}


```

JamesTLopez (commented):
Just some small changes! But overall looks solid

apps/api/src/routes/application-router.ts
surajgoraya: Yeah, maybe something like this?:
> Application can only be rejected if under DAC Review
or
> Application can only be rejected if in DAC Review

apps/api/src/routes/application-router.ts
demariadaniel: Yeah useful information to communicate as part of: https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=95389716&issue=Pan-Canadian-Genome-Library%7Cdaco%7C176

apps/api/src/routes/application-router.ts
```diff
@@ -233,4 +236,73 @@ applicationRouter.post('/applications/approve', jsonParser, async (req, res) =>
 	}
 });
 
+applicationRouter.post('/applications/reject', jsonParser, async (req, res) => {
+	const { applicationId } = req.body;
+
+	if (!applicationId) {
+		res.status(400).json({ message: 'Application ID is required.' });
+	}
+
```
demariadaniel, line 245: Both of these `if` blocks send status `400` and concern `applicationId` so I think it would make sense to combine them:

```
	if (!applicationId || isNaN(parseInt(applicationId))) {
		res.status(400).json({ message: 'Invalid request. ApplicationId is required and must be a valid number.',
			errors: 'MissingOrInvalidParameters', });
	}
```

apps/api/src/routes/application-router.ts
```diff
@@ -233,4 +236,73 @@ applicationRouter.post('/applications/approve', jsonParser, async (req, res) =>
...
+			const message = rejectedApplication.message || 'An unexpected error occurred.';
+			const errors = rejectedApplication.errors;
+			res.status(500).send({ message, errors });
+		}
+	} else {
+		res.status(404).json({ message: 'Application not found.' });
+	}
+});
+
+// Endpoint for reps to request revisions
```
demariadaniel, line 281: Why is this `/request-revisions` endpoint being added? It's not in the ticket and the `requestApplicationRevisions` function does not exist in this branch

apps/api/src/routes/application-router.ts
```diff
@@ -27,8 +27,11 @@ import {
 	getAllApplications,
 	getApplicationById,
 	getApplicationStateTotals,
+	rejectApplication,
+	requestApplicationRevisions,
```
demariadaniel, line 31: This `requestApplicationRevisions` import doesn't exist in this branch, needs to be removed

demariadaniel (changes requested):
Need to clarify what's going on with the `rep-revisions` code

apps/api/src/routes/application-router.ts
```diff
@@ -268,13 +269,40 @@ applicationRouter.post('/applications/reject', jsonParser, async (req, res) => {
...
 	}
 });
 
+// Endpoint for reps to request revisions
+applicationRouter.post('/applications/request-revisions', jsonParser, async (req, res) => {
+	try {
+		const { applicationId } = req.body;
+		const { repId, reviewData, comment, role } = req.body;
+
+		if (!role && (role !== 'REP' || role !== 'DAC')) {
```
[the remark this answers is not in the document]
Azher2Ali: According to the current logic, if role is not null and not undefined then only it will check for comparision with `REP `and `DAC`

apps/api/src/routes/application-router.ts
Azher2Ali: There was conflict between two branches, which is now resolved.

apps/api/src/routes/application-router.ts
Azher2Ali: Current changes, does not include any code related to `requestApplicationRevisions`

demariadaniel (changes requested):
a) `applicationController / rejectApplication` needs a unit test to insure it's functioning properly going forward

b) I'm going to ask we rewrite controller/router/service logic to use the stateManager, I will follow up with details but it's similar to implementation here: https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/apps/api/src/controllers/applicationController.ts#L140

demariadaniel:
@Azher2Ali I've added the Dac Rejected state transition to help set this up

apps/api/src/controllers/applicationController.ts
```diff
@@ -173,3 +173,9 @@ export const approveApplication = async ({ applicationId }: ApproveApplication):
 		return failure(message, error);
 	}
 };
+
+export const rejectApplication = async ({ applicationId }: { applicationId: number }) => {
+	const database = getDbInstance();
+	const service: ApplicationService = applicationSvc(database);
+	return await service.rejectApplication({ applicationId });
+};
```
demariadaniel, line 181: Related to discussion here: https://github.com/Pan-Canadian-Genome-Library/daco/pull/179#pullrequestreview-2607042602

Update `rejectApplication` to follow the example in `approveApplication`

`rejectApplication`	should get application by Id, create new ApplicationStateManager, then call `stateManager.dacReject()` (just added)

`dacReject` covers DB methods of updating the Application record & creating an Action record

apps/api/src/service/applicationService.ts
```diff
@@ -356,6 +357,25 @@ const applicationSvc = (db: PostgresDb) => ({
...
+			if (!application[0]) throw new Error('Application record is undefined');
+
+			return success(application[0]);
+		} catch (exception) {
+			const message = `Error at rejectApplication with applicationId: ${applicationId}.`;
+			logger.error(message);
+			logger.error(exception);
+			return failure(message, exception);
+		}
+	},
```
demariadaniel, line 378: Related to discussion here: https://github.com/Pan-Canadian-Genome-Library/daco/pull/179#pullrequestreview-2607042602

The State Manager handles this DB update, so this service method won't be needed

demariadaniel (approved):
OK looking great, thanks for building this + making updates

Azher2Ali:
> OK looking great, thanks for building this + making updates

Thanks for approving the PR!

Issue #164, opened 2025-01-22
Form Validation - Project Information Page


Buwujiu:
## Descriptions
Implement form validation on the project information page. The validation must ensure:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-3bbffe46-7fff-b896-e2f4-49c69b80fa02"><ul style="margin-top:0;margin-bottom:0;padding-inline-start:48px;"><li dir="ltr" style="list-style-type:disc;font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;" aria-level="1"></li></ul><div dir="ltr" style="margin-left:36pt;" align="left">
Project Title | Required field. Free text field
-- | --
Project Website | Optional field. If provided, UI must perform a basic validation on the URL format.If validation fails, must display an error message “Please enter a valid url. Must begin with http:// or https://.”.
Background | Free text field, users must enter minimum 100 words and maximum 200 words.If the text field is less than 100 or empty, must display an error message “Please enter at least 100 words.”Example: If the text field has more than 200 words, must display an error message “Please enter 200 words or less.”. Example:
Aims | Free text field, users must enter minimum 100 words and maximum 250 words.If the text field is less than 100 or empty, must display an error message “Please enter at least 100 words.”If the text field has more than 250 words, must display an error message “Please enter 250 words or less.”.
Use of Data and Methodology | Free text field, users must enter minimum 100 words and maximum 200 words.If the text field is less than 100 or empty, must display an error message “Please enter at least 100 words.”If the text field has more than 200 words, must display an error message “Please enter 200 words or less.”.
Lay Summary | Same as Use of Data and Methodology.
Relevant Publications | We require the application to submit three publications, of which the applicant is the author or co-author. All three fields are required. UI must perform a basic validation on the URL format.If validation fails, must display an error message “Please enter a valid url. Must begin with http:// or https://.”.

</div></b>
Refer to https://docs.google.com/document/d/1NK5j-ZgwPDwkUmMvC3KfNIdXAhm8JrWmSWBekOtN_SM/edit?tab=t.0

surajgoraya:
[...]

Issue #110, opened 2024-12-16
Application Viewer - Project Information


Buwujiu:
## Descriptions
Build the project information page in viewing mode.
Mockup: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2086-7706&t=kPXzDF7ZRnzz7mFT-4
## Acceptance Criteria
https://docs.google.com/document/d/1NK5j-ZgwPDwkUmMvC3KfNIdXAhm8JrWmSWBekOtN_SM/edit?tab=t.0

patrickdsantos:
Just need to remove the line break at the bottom of the page. Otherwise looks great!

patrickdsantos:
Great work team! No further changes

PR #190, opened 2025-02-03
#110: Application Viewer - Create Project Information Page

surajgoraya:
## Summary

Creates the Project Information in the Application Viewer and integrates the changes in #164. It also fixes various accessibility issues related to forms and form labels. Note that this PR is partially blocked by #161 and #164, and they both must be merged before this one

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/164

### Special Instructions
Please be sure to (re)build the validation library before running: 

```bash
# in the root project folder
pnpm build:all
```

### UI
✨ New / Adds:
- Project Info page (`/pages/applications/sections/project.tsx`), available at `/application/X/project/`
  - Creates a controlled `TextArea` component (`TextAreaBox.tsx`) to be able to type long form free text.
  - Adds the ability to layout form components differently to the global `<Form>` layout with the `labelCol`, `labelAlign` and `layout` props.
     - Used for horizontal layout in publication URL fields
  - Implements the validation schema created in #164, including all translation strings required to make it display errors on the front-end.

🩹 Fixes:
- a11y issues with having nested empty labels when `subLabel` property was not used.
- a11y issues with labels not having a `for` value when a sub label was used. Both labels now use the same for value if both are present, which is valid HTML.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/ui/src/components/pages/application/form-components/InputBox.tsx
```diff
@@ -45,23 +49,63 @@ interface InputBoxProps extends BasicFormFieldProps {
...
-								placeholder={props.placeHolder}
-							/>
+				return props.subLabel ? (
+					<Item
+						label={props.label}
+						required={props.required}
+						name={props.name as string}
+						labelAlign={props.labelAlign ?? undefined}
+						labelCol={props.labelCol ?? undefined}
+						layout={props.layout ?? undefined}
```
JamesTLopez, line 85: labelAlign , labelCol, layout don't seem to be used and doesn't look like its needed. 
Also I think the `props.labelAlign ?? undefined` might be redundant, if the props.labelAlign isn't provided, then it will be undefined. I think the nullish operator can be remove in scenarios like this.

apps/ui/src/components/pages/application/form-components/InputBox.tsx
```diff
@@ -45,23 +49,63 @@ interface InputBoxProps extends BasicFormFieldProps {
...
+						label={props.label}
+						name={props.name as string}
+						required={props.required}
+						rules={[props.rule]}
+						labelAlign={props.labelAlign ?? undefined}
+						labelCol={props.labelCol ?? undefined}
+						layout={props.layout ?? undefined}
+					>
+						{renderControl(field)}
+					</Item>
```
JamesTLopez, line 108: Couple of things:
1. This can be simplified, instead of removing the parent `<Item/>`, conditionally render the child sub `<Item/>`. 
2. Im thinking this might be overkill just to remove a label even though its a11y friendly. I think we can just keep it original and if needed then we can add it back in.

apps/ui/src/components/pages/application/form-components/TextAreaBox.tsx
```diff
@@ -0,0 +1,87 @@
...
+		<Controller
+			name={props.name}
+			control={props.control}
+			render={({ field }) => {
+				return props.subLabel ? (
+					<Item
+						label={props.label}
+						required={props.required}
+						name={props.name as string}
+						labelAlign={props.labelAlign ?? undefined}
```
JamesTLopez, line 57: Same thing here as mentioned in the Inputbox

apps/ui/src/pages/AppRouter.tsx
```diff
@@ -28,6 +28,7 @@ import Introduction from '@/pages/applications/sections/intro';
 import DashboardPage from '@/pages/dashboard';
 import HomePage from '@/pages/index';
 import ManageApplicationsPage from '@/pages/manage/applications';
+import Project from './applications/sections/project';
```
JamesTLopez, line 31: Alias import

apps/ui/src/pages/applications/sections/project.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+import SectionWrapper from '@/components/layouts/SectionWrapper';
+import InputBox from '@/components/pages/application/form-components/InputBox';
+import LabelWithExample from '@/components/pages/application/form-components/labels/LabelWithExample';
+import TextAreaBox from '@/components/pages/application/form-components/TextAreaBox';
+import SectionContent from '@/components/pages/application/SectionContent';
+import SectionFooter from '@/components/pages/application/SectionFooter';
+import SectionTitle from '@/components/pages/application/SectionTitle';
+import { ApplicationOutletContext } from '@/global/types';
+import { useOutletContext } from 'react-router';
```
JamesTLopez, line 35: Import at the top pls.

apps/ui/src/components/pages/application/form-components/InputBox.tsx
surajgoraya: > This can be simplified, instead of removing the parent `<Item/>`, conditionally render the child sub `<Item/>`.

Hmm! Good suggestion, I'll take a look, not sure why I didn't approach it this way initially, will double check.

> Im thinking this might be overkill just to remove a label even though its a11y friendly

Gonna have to disagree here, I know a11y may not be at the trop priority for MVP, but like i18n, doing stuff now as we move along will only be a positive rather than a negative. Not only does it simplify the (eventual) task of adding it explicitly when that becomes a requirement, but it also just makes us more compliment with HTML standard right now. A11y should never be a second thought imo, the more we can do in this phase the better.

Also I should also clarify that this change was required to aid in getting rid of weird visual glitches when in an error state that occur when using the `horizontal` layout (like in the URL publication list)

JamesTLopez: Okie! Then point 1 should be fine

apps/ui/src/i18n/translations.ts
```diff
@@ -71,6 +71,15 @@ const CustomFormErrorTranslationMapping: z.ZodErrorMap = (error, ctx) => {
...
 				return { message: i18n.t('validPostalCode') };
 			}
+			break;
+		case z.ZodIssueCode.custom:
+			switch (error.params?.violation) {
+				case 'tooFewWords':
+					return { message: i18n.t('tooFewWords') };
+				case 'tooManyWords':
+					return { message: i18n.t('tooManyWords') };
+			}
```
JamesTLopez, line 81: Instead of having a switch case, i'm wondering if could we just pass the violation value straight into the translation? Then have a condition to check if the `violation` does not exist and return default error.

apps/ui/src/components/pages/application/form-components/InputBox.tsx
```diff
@@ -17,17 +17,21 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { Form, Input } from 'antd';
-import { Controller, FieldValues, UseControllerProps } from 'react-hook-form';
+import { ColProps, Form, Input } from 'antd';
+import { Controller, ControllerRenderProps, FieldValues, Path, UseControllerProps } from 'react-hook-form';
 
 import { BasicFormFieldProps } from '@/global/types';
+import { FormItemLayout } from 'antd/es/form/Form';
```
JamesTLopez, line 24: Minor, can you move this to the top?

apps/ui/src/components/pages/application/form-components/TextAreaBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { ColProps, Form, Input } from 'antd';
+import { Controller, ControllerRenderProps, FieldValues, Path, UseControllerProps } from 'react-hook-form';
+
+import { BasicFormFieldProps } from '@/global/types';
+import { ReactNode } from 'react';
```
JamesTLopez, line 24: ^

JamesTLopez (changes requested):
Couple of minor changes and a thought!

apps/ui/src/components/pages/application/form-components/InputBox.tsx
surajgoraya: Will do, but I think we need to edit out Prettier config, the position it's at now is auto-formatted

surajgoraya: Or mine might be busted, lol will check

demariadaniel (approved):

JamesTLopez (approved):
I YIELD ⚔️

Issue #188, opened 2025-02-03
Form Validation - Ethics


Buwujiu:
## Descriptions
Implement the form validation on ethics page. 
## Acceptance Criteria
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-43df0ded-7fff-0509-410a-bfcb9706b61a"><div dir="ltr" style="margin-left:0pt;" align="left">
Form Validation Rule | Either an exemption letter or ethics approval letter must be uploaded. Otherwise users should not be allowed to click on Next.
-- | --

</div></b>

Buwujiu:
Ethics UI and validation is done.

Issue #184, opened 2025-02-03
Application Viewer - Ethics


Buwujiu:
## Descriptions
Implement the Ethics UI.
Mockup: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2087-6364&t=qY5lykj73deJHvHO-4
## Acceptance Criteria
Sepcs: https://docs.google.com/document/d/1NK5j-ZgwPDwkUmMvC3KfNIdXAhm8JrWmSWBekOtN_SM/edit?tab=t.0

JamesTLopez:
### Feedback round 1
Feedback addressed: https://github.com/Pan-Canadian-Genome-Library/daco/pull/264

patrickdsantos:
Perfect! Thanks team, no further changes.

PR #198, opened 2025-02-07
#184: Ethics Section UI and Validation

JamesTLopez:
## Summary

UI Implementation of the Ethics Section of the Application Wizard and basic validation setup.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/184
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/188

## Description of Changes

- create Ethics page
- create BlockRadioBox
   - radio box with a grey background
-  create basic zod validation for the file upload and the type of file to be uploaded
- custom rhf integration for file upload
- error messages for upload file specific errors (only client side)
   - notification for improper file type
   - notification for file size too big 

### Special Instructions
Before running these changes, you will need to rebuild the the validation package.
Please run the following in the `apps` or `daco` folder:
```
pnpm build:all
```

## Readiness Checklist

- [ ] **Self Review**
- [ ] **PR Format**
- [ ] **Labels Added**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation

### DIFF
--- a/apps/ui/src/components/pages/application/form-components/BlockRadioBox.tsx
+++ b/apps/ui/src/components/pages/application/form-components/BlockRadioBox.tsx
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2025 The Ontario Institute for Cancer Research. All rights reserved
+ *
+ * This program and the accompanying materials are made available under the terms of
+ * the GNU Affero General Public License v3.0. You should have received a copy of the
+ * GNU Affero General Public License along with this program.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { ColProps, Flex, Form, Radio, Row, theme } from 'antd';
+import { FormItemLayout } from 'antd/es/form/Form';
+import { Controller, ControllerRenderProps, FieldValues, Path, UseControllerProps } from 'react-hook-form';
+
+import { pcglColors } from '@/components/providers/ThemeProvider';
+import { BasicFormFieldProps } from '@/global/types';
+import { EthicsFileTypes } from '@pcgl-daco/validation';
+
+const { Item } = Form;
+const { useToken } = theme;
+
+interface CheckboxValues {
+	value: EthicsFileTypes;
+	label: string | React.ReactElement;
+}
+
+interface InputBoxProps extends BasicFormFieldProps {
+	subLabel?: string;
+	placeHolder?: string;
+	labelAlign?: 'left' | 'right';
+	labelCol?: ColProps;
+	layout?: FormItemLayout;
+	options: CheckboxValues[];
+}
+
+const BlockRadioBox = <T extends FieldValues>(props: UseControllerProps<T> & InputBoxProps) => {
+	const { token } = useToken();
+	/**
+	 * Renders the input box component. We currently use two nested labels to display certain information on the
+	 * application form (see Applicant Info). But in the case where only one label is needed, we should only render
+	 * one label to avoid visual and a11y issues.
+	 *
+	 * This function helps reduce redefining the Input field.
+	 * @param field the field attribute from `react-hook-from`
+	 * @returns `ReactNode` with the input component.
+	 */
+	const renderControl = (field: ControllerRenderProps<T, Path<T>>) => {
+		return (
+			<Radio.Group {...field} disabled={props.disabled}>
+				<Row justify="start" gutter={[0, 10]}>
+					{props.options.map((item) => {
+						return (
+							<Flex
+								key={item.value}
+								style={{
+									backgroundColor: pcglColors.greyLight,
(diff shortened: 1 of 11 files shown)


apps/ui/src/pages/applications/sections/ethics.tsx
```diff
@@ -0,0 +1,141 @@
...
+const MAX_FILE_SIZE = 5000000;
+
+const Ethics = () => {
+	const { t: translate } = useTranslation();
+	const { isEditMode } = useOutletContext<ApplicationOutletContext>();
+	const { control, watch, setValue } = useForm<EthicsSchemaType>({});
+	const value = watch('ethicsApproval');
+	const { token } = useToken();
+
+	const uploadFile: UploadProps = {
```
JamesTLopez, line 57: I pull the props/attributes into a variable then spread it on the `<Upload {...uploadFile}/>` because it seemed like there would be a lot of worked being done here, specifically on `onChange`  and `beforeUpload`. 

The alternative here is we create two specific functions for `onChange` and `beforeUpload` and the rest goes into the actual component attributes. Im torn on what is nicer, if theres any suggestions one what looks nicer here, im open to ideas!

apps/ui/src/pages/applications/sections/ethics.tsx
```diff
@@ -0,0 +1,141 @@
...
+				return false;
+			}
+		},
+		onChange: (info) => {
+			// Add file path to the rhf here, once the file upload is complete. Will probably change once fileupload functionality is implemented
+			if (info.file.status === 'done') {
+				setValue('uploadPath', 'successful path');
+			} else {
+				setValue('uploadPath', 'failed path');
+			}
```
JamesTLopez, line 88: I am not using a `<Controller/>` pattern here. Instead I opt'd to use the the `setValue` to add the Upload related information to the form as the steps to use the controller makes the component much more complex. This lets us to be more custom with where we want to add the upload file related data.

packages/validation/src/schemas.ts
```diff
@@ -83,3 +85,8 @@ export const projectInformationSchema = z.object({
 	relevantPublicationURL2: NonEmptyString.url(),
 	relevantPublicationURL3: NonEmptyString.url(),
 });
+
+export const ethicsSchema = z.object({
+	ethicsApproval: EthicsEnum,
+	uploadPath: NonEmptyString,
```
JamesTLopez, line 91: I am unsure of what the meta data looks like for uploadPath when we push it up to the backend here. This is temporary until we know what methods we are using to upload files. At the very least, we need to know if the file is either `Exemption` or `EthicsLetter`.

packages/validation/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+export enum EthicsFileEnum {
+	EXEMPTION = 'exemption',
+	ETHICS_LETTER = 'ethicsLetter',
+}
```
joneubank, line 23: No changes for this PR, but I am going to recommend a refactor of our work in this project to remove the use of `enums`. I'm leaving this link here as a good summary of what some of the concerns with `enums` are: https://www.totaltypescript.com/why-i-dont-like-typescript-enums

Note that they way you defined the enum is the best way, with named string values for every enum key. This won't expose us to any strange type issues down the line, but it still feels weird to rely on TS's generated JS code.

packages/validation/src/schemas.ts
joneubank: I don't think we need this path. The ethics file really just needs the file uploaded. The only meta data we store is if the ethics letter is required.

joneubank (changes requested):
Good stuff. Couple simple changes to make.

- removing upload path
- removing invalid type assertions

apps/ui/src/components/pages/application/form-components/BlockRadioBox.tsx
```diff
@@ -0,0 +1,116 @@
...
+	return (
+		<Controller
+			name={props.name}
+			control={props.control}
+			render={({ field }) => {
+				return (
+					<Item
+						label={props.label}
+						required={props.required}
+						name={props.name as string}
```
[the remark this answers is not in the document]
JamesTLopez: This actually something I encountered initially when making the custom component, and theres a reason why I ultimately decided and think its fine for this particular scenario.

There is a trade off for the changing the `props.name to string` instead of its default type `TName`. 
Heres a custom type I tried to resolve this issue as a reference:
```js
export type UseCustomControllerProps<
	TFieldValues extends FieldValues = FieldValues,
	TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
> = {
	name: string; // Originally here, it would be type `TName` 
	rules?: Omit<RegisterOptions<TFieldValues, TName>, 'valueAsNumber' | 'valueAsDate' | 'setValueAs' | 'disabled'>;
	shouldUnregister?: boolean;
	defaultValue?: FieldPathValue<TFieldValues, TName>;
	control?: Control<TFieldValues>;
	disabled?: boolean;
};

```

Firstly, if we change the `name` to type `string`, it solves that type assertion issue, but another type issue then occurs on the `name` field of the `<Controller/>` component:

``` js
// This is the hook version for simplicity
	const { field } = useController({
		name: props.name as Path<T>, // Type 'string' is not assignable to type 'Path<T>
		control: props.control,
		rules: { required: true },
	});

```
To resolve this, I would have to do something similar like `as Path<T>`. 

The main reason why I think this is type assertion is fine in this scenario is because typescript actually complains if you do not use the correct fields as defined in the zod Schema. For example take a look at the `name` property  type of any of the custom components on `Institutional Rep or Applicant` pages:

image

As you can see from the example, the component will expect any one of the fields defined in the zodschema, and if passed any other string it would then complain. 

Ultimately, if I changed the `props.name` to `string` and not `TName`, we would lose the zod-schema expected values on the name property and we would have to type assert ` props.name as Path<T>,` anyways.

If theres a typescript way of supporting both of these that im not seeing, I am open to suggestions!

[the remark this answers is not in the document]
JamesTLopez: NOTE: 
After discussion with Jon, to ensure that the passed value to the name attribute for component `<Item/>` is to apply a runtime conversion as `const name = ${props.name}` or `${props.name}` in place of the attribute.

apps/ui/src/components/pages/application/form-components/BlockRadioBox.tsx
```diff
@@ -85,7 +85,7 @@ const BlockRadioBox = <T extends FieldValues>(props: UseControllerProps<T> & Inp
 					<Item
 						label={props.label}
 						required={props.required}
-						name={props.name as string}
+						name={`${props.name}`}
```
joneubank, line 88: 👍

JamesTLopez:
> Thoughts on how to make the number here configurable based on the field? I don't think we need to accommodate multiple maximums but we are ending up in a situation where the validation rule and the number will need to be maintained in multiple places. We'll also need to keep the number consistent between each language in the translation. Would be better if the specific number could be passed in as an argument.

Solution to this will be referred to the following ticket https://github.com/Pan-Canadian-Genome-Library/daco/issues/203

Issue #176, opened 2025-01-27
API Central Error Handling


demariadaniel:
## Descriptions
Implement error response middleware so all endpoints have coverage for standard 400/500 status errors

Similar to Lyric example: 
https://github.com/overture-stack/lyric/blob/main/packages/data-provider/src/middleware/errorHandler.ts

Dan and Azher to work together on this.

surajgoraya:
Related #152 -- I will be adding in OHCRN's schema validation package (with changes)

Buwujiu:
We have models to borrow from.

Buwujiu:
Have not started working, move to sprint 9.

Buwujiu:
Dan approved the PR, but has a few things to confirm with Jon.

Issue #152, opened 2025-01-15
QA Feedback on Edit Application Endpoint


Buwujiu:
## Descriptions
1. When testing edit app endpoint using new or existing application, if a non-existent field is added to the payload, 
for example field "chocolate", API returns 200, and field is not added. 
This is a confusing behaviour, it could throw an error to forbid adding non-existent fields. 

2. Tested the same endpoint using new or existing application, edit query to only contain "id" and no "update", 
API returns 200, and the application object is not updated. 
This is confusing, it should not allow edit when "update" is missing.

edsu7:
Bugs addressed but see additional observations.

## Descriptions
> 1. When testing edit app endpoint using new or existing application, if a non-existent field is added to the payload, 
for example field "chocolate", API returns 200, and field is not added. 
This is a confusing behaviour, it could throw an error to forbid adding non-existent fields. 

![Image](https://github.com/user-attachments/assets/c21dad47-d5a0-42bc-86a0-1065f17d8ab1)

Error now successful displayed

> 2. Tested the same endpoint using new or existing application, edit query to only contain "id" and no "update", 
API returns 200, and the application object is not updated. 
This is confusing, it should not allow edit when "update" is missing.

![Image](https://github.com/user-attachments/assets/3c9e236a-d46b-453a-9064-5ca3ee3b76d1)

Error now successful displayed

## Additional observations:

Edit still shows camel case and only accept camel case

![Image](https://github.com/user-attachments/assets/7c223981-2755-4bad-a108-44cd27fb44ce)

![Image](https://github.com/user-attachments/assets/4c1dbabf-fc42-4f9f-8b8b-e84a998e3c0e)

![Image](https://github.com/user-attachments/assets/190f975c-b03c-4932-9740-dfb750114ac3)

Buwujiu:
Tested in Dev, working as expected, closing!

PR #202, opened 2025-02-11
#152: Address QA Feedback - Edit Application Endpoint

surajgoraya:
## Summary
Addresses feedback QA feedback received with regards to the `/applications/edit` endpoint by making use of our shared `@pcgl-daco/validation` package to check the structure of the incoming JSON. 

To do this in a reusable fashion, `@pcgl-daco/request-utils` was added. It is based off of the `express-request-validation` package found in OCHRN. This package may also serve as a place to place centralized error handling middleware once implemented (see: https://github.com/Pan-Canadian-Genome-Library/daco/issues/176) 

Related: [OHCRN/platform/packages/`express-request-validation`](https://github.com/OHCRN/platform/tree/develop/packages/express-request-validation).

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/152
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/176 (Related)

### Special Instructions
Before running these changes, you will need to build and install `@pcgl-daco/request-utils`:
```bash
# in the root project folder
pnpm i && pnpm build:all
```

### Server
- Adds `withSchemaValidation` middleware to the `/applications/edit` endpoint to check the incoming JSON object against our Zod validation schema.
- Adds `apiZodErrorMapping` in `/utils/validation.ts` to provide useful messages to the end user for certain Zod errors (custom refine/superRefine, required properties in objects).

### `@pcgl-daco/request-utils`
- Adds a new package: `@pcgl-daco/request-utils` based off of OHCRN's `express-request-validation` package.
   - This package can serve as a starting point for other middleware related to validation such as global error handling.
- **🚚  Changes from `express-request-validation`** :
  - Various adaptations to `express-request-validation` code to fit our use case (types, splitting up files, etc..). 
  - Introduces the ability to add a `ZodErrorMap` to the schema validation function allowing us to remap error strings as required.
  - Renames `withRequestValidation` to `withSchemaValidation` for clarity.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
  - image

- [x] Documentation


surajgoraya:
Approval criteria based off ticket:
1. Invalid in update object:
image

2. No update object:
image

joneubank:
Special instructions probably need to be in the opposite order (install before build) ;)
```sh
# in the root project folder
pnpm build:all && pnpm i
```

packages/validation/src/schemas.ts
```diff
@@ -94,3 +95,44 @@ export const ethicsSchema = z.object({
...
+			institution_postal_code: NonEmptyString,
+			institution_building: NonEmptyString,
+			project_title: NonEmptyString,
+			project_website: NonEmptyString.url(),
+			project_abstract: NonEmptyString,
+			project_methodology: NonEmptyString,
+			project_summary: NonEmptyString,
+		})
+		.partial()
+		.strict()
```
joneubank, line 134: No harm in this `strict()` restriction, but whats the benefit/motivation for this?

packages/validation/src/schemas.ts
```diff
@@ -94,3 +95,44 @@ export const ethicsSchema = z.object({
 export const requestedStudySchema = z.object({
 	requestedStudy: NonEmptyString,
 });
+
+export const applicationEditSchema = z.object({
```
joneubank, line 99: I would like to use this moment to start to add some more organization to this every growing file.

This new schema is specifically the schema for a Request. The UI will use the type to ensure they are sending the correct data, and the API will use the schema to validate the incoming request body.

Similarly, Response schemas will be needed from these endpoints, as they are used by the API to declare the return type of the route and are used by the UI to parse the response.

Can we create a separate directory called routes and create one file for each router in the api, the content of that file should be the request and response schemas (plus inferred types) for all the endpoints in that router.

So this schema would move to the file `/src/routes/applicationRoutes.ts` and should be renamed `editApplicationRequestSchema`.

packages/validation/src/schemas.ts
```diff
@@ -26,6 +26,7 @@ export type ApplicantInformationSchemaType = z.infer<typeof applicantInformation
 export type InstitutionalRepSchemaType = z.infer<typeof institutionalRepSchema>;
 export type ProjectInformationSchemaType = z.infer<typeof projectInformationSchema>;
 export type CollaboratorsSchemaType = z.infer<typeof collaboratorsSchema>;
+export type ApplicationEditSchemaType = z.infer<typeof applicationEditSchema>;
```
joneubank, line 29: Types do not need the word `Type` in their name... they are always a type.

This is also not the type of the schema, it is the type of the object defined by the schema.

This should be named `EditApplicationRequest`. See other comment about refactor for moving this into a file dedicated to applicationRouter requests/responses.

joneubank (changes requested):
Please add a README.md for the new package, and update the root level readme with information about the added package in the section about monorepo structure.

packages/validation/src/schemas.ts
joneubank: Also, it is confusing to have type declarations so far fro the schemas that they are inferred from. Usually I would declare the type on the line following the zod schema.

packages/validation/src/schemas.ts
surajgoraya: Currently using this `strict()` to make sure that no unknown keys are sent into the update object as specified by the QA feedback, we could get rid of it, it just means that unknown keys won't be rejected at the zod level.

packages/validation/src/schemas.ts
surajgoraya: I've gone ahead and made these changes! I've been scratching my head trying to figure out if theres more ways to break down `schemas.ts`? It still feels like quite a large file, and so far, all the schemas inside there are specifically for the application viewer / editor

packages/validation/src/routes/applicationRoutes.ts
```diff
@@ -0,0 +1,63 @@
...
+			applicant_first_name: NonEmptyString,
+			applicant_middle_name: EmptyOrOptionalString,
+			applicant_last_name: NonEmptyString,
+			applicant_title: NonEmptyString,
+			applicant_suffix: EmptyOrOptionalString,
+			applicant_position_title: NonEmptyString,
+			applicant_primary_affiliation: NonEmptyString,
+			applicant_institutional_email: NonEmptyString.email(),
+			applicant_profile_url: NonEmptyString.url(),
+			institutional_rep_title: NonEmptyString,
```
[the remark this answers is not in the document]
surajgoraya: Whoops! Good points! Partial updates are allowed, the `partial()` in the Zod schema allows this. I'll take a look at the schema again!

Honestly, Ideally we should be using the already created validations for the frontend here as well maybe combining them somehow? But as of right now since we have different variable casing from the front end and expected by the backend we can't just yet.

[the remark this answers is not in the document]
demariadaniel: Thanks I missed partial() there. I am intruigued how we could make this better, will take a look.

packages/validation/src/routes/applicationRoutes.ts
```diff
@@ -0,0 +1,64 @@
...
+			project_website: OptionalURLString,
+			project_abstract: ConciseWordCountString,
+			project_methodology: ConciseWordCountString,
+			project_summary: ConciseWordCountString,
+		})
+		.partial()
+		.strict()
+		.refine((updateObj) => Object.keys(updateObj).length !== 0, {
+			params: { violation: 'noEmptyObject' },
+		}),
```
joneubank, line 63: Don't know if this is necessary. Sure the request would be useless with an empty object but I don't know that it would be invalid.

packages/validation/src/schemas.ts
```diff
@@ -90,14 +85,17 @@ export const projectInformationSchema = z.object({
...
 	ethicsApproval: EthicsEnum,
 });
 
+export type RequestedStudySchemaType = z.infer<typeof requestedStudySchema>;
 export const requestedStudySchema = z.object({
 	requestedStudy: NonEmptyString,
 });
 
+export type AgreementsSchemaType = z.infer<typeof agreementsSchema>;
 export const agreementsSchema = z.object({
```
joneubank, line 99: Using the const before it is defined. TS is happy with this but I find this logically confusing. Can we put the types after the schema?

joneubank (dismissed):
Good stuff. Minor adjustments suggested but I am approving so we can all start using this middleware.

packages/validation/src/routes/applicationRoutes.ts
surajgoraya: Hmm, yeah, in a way there's not any harm letting the user pass in an empty object, however, at the same time, I think it would probably be a good idea to let the user know, since a "update" to nothing doesn't make any sense, plus, it was a part of the ticket that this PR is addressing so I think it still makes sense to add it imo!

JamesTLopez (dismissed):

Issue #50, opened 2024-11-14
Document the Auth Plan


Buwujiu:
Write a technical document on Auth Integration. 
Based on the user role and permissions. 
https://editor.swagger.io/?url=https://raw.githubusercontent.com/CanDIG/pcgl-authz/refs/heads/main/authz_openapi.yaml

Buwujiu:
Jon will finish OHCRN session mgnt first, then bring the model to DACO.

Buwujiu:
Jon has been focusing on publishing the Song and Score new versions, will do this in sprint 6.

Buwujiu:
Document is complete, closing.

joneubank:
https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/docs/auth-setup.md

PR #220, opened 2025-02-15
#50: Authentication Implementation

joneubank:
## Summary

Adds User authentication functionality relying on an OIDC provider.

This adds the ability to login and logout a user.

A doc file is added to describe how the auth flow works and how developers can interact with it.

### Note on User Authorization

At the current time there is no source for user permissions available. There is a plan for PCGL Auth service currently in development, but this is not integrated with this work. As a result, this PR handles integration with the OIDC service but does not provide a mechanism to differentiate a user between an Applicant and a DAC Member. This will be added in later tickets.

### Related Issues

- #50 

### Docker Compose
- Adds Valkey server for session storage

### UI
- Creates `UserProvider` that will track information about the user
  - user logged-in status
  - user role, for use with protected routes
  - checks user session on page load (gets triggered after log in and log out)
- Adds `ProtectedRoute` component that can check if user is authorized to visit a given route, and redirect if they are not.
- Enables Login/Logout through Header button and Apply for Access modal
- Optional UI environment variable to disable all auth checks, allowing UI editing without being logged in

### API
- Adds session management
  - Uses `express-session` to manage session cookie and make session data available in express requests
  - Uses Valkey for session data storage
- Adds `authRouter` to handle login and logout flow
  - /auth/user endpoint for UI to inspect user info from current session

### Developer Instructions
- Update your daco-api .env file
  - Valkey and session config variables are required to start API
  - Auth config variables are required to use the auth features
  - Ask @joneubank for auth client credentials for test environment
- Install new npm dependencies `pnpm i`
- Re-run `docker-compose up` to start Valkey

### Deployment Instructions
Before merging or deploying these update, the environment variables for environments running the built docker image will need to have the Auth and Valkey configs added. When running in a docker container the NODE_ENV variable is set to be production. There is a coded rule that requires the env config to have auth enabled in production, so the updated configs are required for the server to startup when running from the docker image.

## Follow Up
There are some related functionalities that are not included here in order to limit the scope of this already large PR.

- API Auth middleware and other auth checks on existing routes
- UI Automatically updating user state when session expires
- UI communicating auth failures or redirecting when accessing applications that are not found or belong to another user

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [x] Documentation


apps/api/src/session/index.ts
```diff
@@ -0,0 +1,48 @@
...
+	refreshToken: string;
+	refreshTokenIat: number;
+};
+
+declare module 'express-session' {
+	interface SessionData {
+		user: SessionUser;
+		account: SessionAccount;
+	}
+}
```
joneubank, line 42: Declare session data structure here.

apps/api/.env.schema
```diff
@@ -1,9 +1,24 @@
-# Server
-PORT=3000
-IS_PROD=false
+# -- Server --
+# PORT=3000
+SESSION_KEYS=newsecret,oldsecret
+# SESSION_MAX_AGE=1800000
```
demariadaniel, line 4: Why are some of these commented out?

apps/api/src/server.ts
```diff
@@ -38,17 +39,21 @@ const API_DOCS_PATH = `api-docs`;
 const startServer = async () => {
 	const app = express();
 
+	// app.use(ExpressLogger({ logger, excludeURLs: ['/auth/token'] }));
 	app.use(ExpressLogger({ logger }));
 
+	app.use(sessionMiddleware);
```
joneubank, line 45: Adding session middleware here. It is defined in `./session/sessionMiddleware.ts` and uses the valkeyClient defined in `./session/valkeyClient.ts`

apps/api/src/routes/authRouter.ts
```diff
@@ -0,0 +1,239 @@
...
+			idToken,
+			refreshToken,
+			refreshTokenIat,
+		};
+		req.session.user = {
+			userId: sub,
+			familyName,
+			givenName,
+		};
+		req.session.save();
```
joneubank, line 196: This is where user session info is set. Happens after successfully confirming a login.

apps/api/README.md
```diff
@@ -23,13 +23,19 @@ Follow these steps to prepare your local machine to run this application from so
...
+| `VALKEY_HOST`     | Valkey host address.                                                                                                                                                                                                                                                                                                                       | `string` | Required |            |
+| `VALKEY_PASSWORD` | Valkey user password.                                                                                                                                                                                                                                                                                                                      | `string` | Required |            |
+| `VALKEY_USER`     | Valkey user name.                                                                                                                                                                                                                                                                                                                          | `string` | Required |            |
```
demariadaniel, line 38: Auth values are not documented `AUTH_CLIENT_ID` etc

apps/api/src/routes/authRouter.ts
```diff
@@ -0,0 +1,239 @@
...
+			method: 'POST',
+			headers: {
+				Authorization: `Bearer ${account.accessToken}`,
+				'content-type': 'application/x-www-form-urlencoded',
+			},
+			params,
+		});
+
+		// On logout success we can clear the session data.
+		resetSession(req.session);
```
joneubank, line 84: Session information is cleared on successful logout.

apps/ui/src/components/providers/UserProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+type UserState = {
+	isLoading: boolean;
+	isLoggedIn: boolean;
+	refresh: () => void;
+} & Partial<UserResponse>;
+
+const UserContext = createContext<UserState>({ isLoading: true, isLoggedIn: false, refresh: () => {} });
+
+export function UserProvider({ children }: PropsWithChildren) {
+	const { data, isLoading, refetch } = useGetUser();
```
joneubank, line 36: Fetches user data on page load.

Big todo is to refresh this if the app has not been reloaded in a while, or if a request from the user comes back as unauthorized.

apps/api/.env.schema
joneubank: To show that the env variable is available if you want to change it, but doesn't need to be set by default. The value shown is the default value.

apps/api/src/config/serverConfig.ts
```diff
@@ -18,17 +18,30 @@
...
 
 import { z } from 'zod';
+
 import EnvironmentConfigError from './EnvironmentConfigError.js';
 
 const serverConfigSchema = z.object({
-	PORT: z.coerce.number().optional().default(3000),
 	NODE_ENV: z.string().optional().default('development'),
 	npm_package_version: z.string().optional().default('unknown'),
+
```
JamesTLopez, line 27: I think we can remove this line break here

JamesTLopez (commented):
Minor stuff but looks good and clean overall! Ill take another glace and test more at another time.

apps/api/src/routes/authRouter.ts
```diff
@@ -0,0 +1,239 @@
...
+	refresh_token_iat: zod.number(),
+	id_token: zod.string(),
+});
+
+const oidcUserinfoResponseSchema = zod.object({
+	sub: zod.string(),
+	given_name: zod.string().optional(),
+	family_name: zod.string().optional(),
+	email: zod.string().optional(),
+});
```
surajgoraya, line 116: Would we benefit from dropping these into the shared validation lib at all? It'd be nice to have all the Zod validation centralized in one place 🤔

apps/api/src/utils/buildQueryParams.ts
```diff
@@ -0,0 +1,17 @@
...
+ * const params = { a: 1, b: false, c: undefined, d: 'two' };
+ * buildQueryParams(params); // ?a=1&b=false&d=two
+ */
+export function buildQueryParams(params: Record<string, string | number | boolean | undefined>): string {
+	const paramString = Object.entries(params)
+		.filter(([key, value]) => value !== undefined)
+		.map(([key, value]) => `${key}=${value}`)
+		.join(`&`);
+	return `?${paramString}`;
+}
```
ciaranschutte, line 17: totally fine. this API also exists https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

apps/api/src/routes/authRouter.ts
```diff
@@ -0,0 +1,239 @@
+import axios, { AxiosError } from 'axios';
```
demariadaniel, line 1: Copyright

apps/ui/src/api/useGetUser.ts
```diff
@@ -0,0 +1,56 @@
...
+				};
+
+				switch (response.status) {
+					case 404:
+						error.message = translate('errors.http.404.title');
+						error.errors = translate('errors.http.404.message');
+						break;
+				}
+
+				throw error;
```
surajgoraya, line 48: The API currently seems to return a `400`, would it be a good idea to handle that HTTP status here in the switch case? Right now the only scenario seems to be regarding the Auth being disabled, so maybe we can skip it for now.

apps/ui/src/components/Header.tsx
```diff
@@ -150,7 +181,7 @@ const HeaderComponent = () => {
 						<Link
 							key={`menuItem-${key}`}
 							style={linkStyle}
-							target={menuItem.target ?? '_blank'}
+							// target={menuItem.target ?? '_blank'}
```
demariadaniel, line 184: Unused commented line

apps/ui/src/components/ProtectedRoute.tsx
```diff
@@ -0,0 +1,82 @@
...
+		return <Redirect />;
+	}
+
+	if (requiredRoles) {
+		if (!role) {
+			return <Redirect />;
+		}
+		if (!requiredRoles.includes(role)) {
+			return <Redirect />;
+		}
```
demariadaniel, line 76: Could combine these two `if (!role || !requiredRoles.includes(role))`

apps/ui/src/components/Header.tsx
```diff
@@ -76,16 +80,47 @@ const HeaderComponent = () => {
...
+	/**
+	 * Default action when a button in the menu is clicked, used particularly for the mobile menu which should close after click.
+	 * @param buttonAction The function for the action needed to be performed.
+	 */
+	// const onMenuButtonClick = (buttonAction: VoidFunction) => {
+	// 	if (isResponsiveMode) {
+	// 		setResponsiveMenuOpen(false);
+	// 	}
+	// 	buttonAction();
+	// };
```
demariadaniel, line 92: Maybe a merge issue? This section duplicates lines 72-81

apps/api/.env.schema
```diff
@@ -1,9 +1,24 @@
...
+UI_HOST=https://localhost:5173
```
demariadaniel, line 5: This should be `http://` we don't use https locally

apps/api/src/config/authConfig.ts
```diff
@@ -0,0 +1,57 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { z } from 'zod';
+import EnvironmentConfigError from './EnvironmentConfigError.js';
+import { serverConfig } from './serverConfig.js';
```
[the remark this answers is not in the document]
joneubank: Files in same level I'll use the relative import. I'll switch to absolute for anything that is up one level or more.

apps/api/src/config/serverConfig.ts
joneubank: Added comments to explain the separate sections.

apps/api/src/utils/buildQueryParams.ts
joneubank: I used this at some point, I'm trying now to remember why I thought this was necessary...

joneubank: Can't remember why. Removing the custom solution to use the built in api.

apps/api/src/routes/authRouter.ts
joneubank: In this case, I don't think there is value to that. These are responses from an external service and it is unlikely they'll be reused.

I could see moving them into a file dedicated to an OIDC client, especially if we need to support implementations for different auth services, but atm I think this is only useful within this file.

apps/ui/src/api/useGetUser.ts
joneubank: This looks like copy/paste from another fetch hook, so I'll update it to match the expected results.

apps/ui/src/components/ProtectedRoute.tsx
joneubank: Agreed this is equivalent. I separated them to clearly show the two cases that need handling. I think I'll leave them separate for now, indicating they could be handled differently.

joneubank:
Feedback comments applied. I also simplified the config. Auth is now enabled by default but can be disabled with `DISABLE_AUTH=true`. The redirect paths on login and logout are now set in code constants instead of being configurable, since this is preset application logic and not varied between environments.

Issue #57, opened 2024-11-19
DACO OIDC Login Flow


Buwujiu:
Design the Login flow, draft a plan for UI-Server session management. 
Jon will work with Dan and Leo to produce a flow diagram and documents.

Buwujiu:
2 draft PRs are ready, will be up for review on Thursday.

Buwujiu:
Tested in dev, was able to 1. register using OICR email and create a password. 2. login using OICR email. 3. Logout works as expected.

PR #221, opened 2025-02-16
#50: API Route Authentication Protections

joneubank:
#50 and #57: Implement Auth plan on API

## Summary

Add authorization handling to all endpoints in the API.

This leverages the user session which will indicate if the user is logged in and what role they have. An Express middleware function is added that will prevent a handler from running based on the user's role, defined in their session. Additional checks need to be added inside of some endpoints to do checks, example: ensuring a user editing an application actually owns that application

### Related Issues

- #50 
- #57 

### API
- Add authMiddleware() and additional user checks to all routes
- Add a `ResponseWithData<DataType, ErrorCases>` type that allows the express handler to validate return types
- Standardize error responses to have a standard structure using `ResponseWithData` for all routes
- Refactor `Result<Data, ErrorCaes>` type to specify error cases
- Use updated result type on all Controller and Service responses so that we safely handle all outcomes in the routers
- Apply replace `bodyParser` with  Express's built in json parser. Use this handler for all routes, not applied per endpoint.

### UI
- Update some type aliases names that were updated for clarity, should have no functional changes

### Validation
- Clean up generic types for handler in body validation so that the types are persisted
- The generic types in the query and param validators are not persisting types still, I reccomend we remove these in a future PR in favour of pattern that validates them within the handler.
- Added some shared DTO types to clarify the returns from a few endpoints

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [ ] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


joneubank:
This has many changes due to the refactor of the `Result` type and the introduction of the `ResponseWithData` type. These were added to standardize how we communicate errors from functions and the formatting of errors in all endpoints. 

The functional addition is the `authMiddleware()` which checks that the user is not anonymous, or has a specific role if specified. Most routes have additional auth rules that require interaction with the database, primarily to check if the user is the creator of the application being edited. These couldn't easily be handled in middleware so I wrote a function to help standardize this. There is a lot of error case handling, but that's cause there are a lot of place an error could occur. I would be open to ways to simplify this, as right now there is a lot of repeated code throughout.

Also, I need to update the swagger before this should be approved.

apps/api/src/external/oidcAuthClient.ts
```diff
@@ -0,0 +1,111 @@
...
+import { type AuthConfig } from '@/config/authConfig.js';
+import BaseLogger from '@/logger.js';
+import { failure, success, type AsyncResult } from '@/utils/results.js';
+
+const logger = BaseLogger.forModule('oidcAuthClient');
+
+/*
+ * OIDC Authorization Flow Step 1
+ * This is a redirection to the OIDC provider /authorize endpoint
+ *  */
```
ciaranschutte, line 33: `/**` for tsdoc?

apps/ui/src/providers/Providers.tsx
```diff
@@ -28,9 +28,11 @@ const Providers = ({ children }: { children: React.ReactNode }) => {
...
 		<QueryClientProvider client={queryClient}>
 			<ThemeProvider>
-				<BrowserRouter>
-					<UserProvider>{children}</UserProvider>
-				</BrowserRouter>
+				<UserProvider>
+					<BrowserRouter>
+						<UserProvider>{children}</UserProvider>
+					</BrowserRouter>
+				</UserProvider>
```
ciaranschutte, line 35: double `UserProvider`?

packages/request-utils/src/validation.ts
```diff
@@ -135,12 +135,12 @@ function withParamsSchemaValidation<ReqParams>(
...
  */
-function withQuerySchemaValidation<ReqQuery>(
+function withQuerySchemaValidation<ReqQuery extends object>(
 	querySchema: ZodSchema<ReqQuery>,
 	zodErrorMapping: ZodErrorMap | undefined,
-	handler: RequestHandler<ParamsDictionary, any, any, qs.ParsedQs>,
-): RequestHandler {
-	return async (request: Request, response: Response, next: NextFunction) => {
+	handler: RequestHandler<ParamsDictionary, any, any, ReqQuery>,
+): RequestHandler<ParamsDictionary, any, any, ReqQuery> {
```
ciaranschutte, line 142: `any`, are you ok? 🕺

JamesTLopez (commented):
Tested locally, went through the entire PR and focused primarily on my stuff.
Some small changes/organization, but there is some type issues on the frontend with the `CollaboratorsResponse` changes which shouldn't effect functionality. Other than that solid stuff! Excited to have this in 🔥

apps/api/src/controllers/applicationController.ts
```diff
@@ -230,11 +248,12 @@ export const submitRevision = async ({ applicationId }: { applicationId: number
 
 		const appStateManager = new ApplicationStateManager(application);
 
+		appStateManager.can(ApplicationStateEvents.submit_dac_revisions);
```
demariadaniel, line 251: Should this be a variable or in the conditional? Just calling `.can`?

apps/api/src/external/oidcAuthClient.ts
```diff
@@ -0,0 +1,111 @@
...
+	};
+
+	try {
+		const tokenResponse = await axios.get(oauth2TokenUrl, { params });
+		const parsedTokenResponse = oidcTokenResponseSchema.safeParse(tokenResponse.data);
+		if (!parsedTokenResponse.success) {
+			logger.error(`OIDC Provider returned tokens with unexpected format`, parsedTokenResponse.error.message);
+			return failure('SYSTEM_ERROR', `Token response has unexpected format.`);
+		}
+		return success(parsedTokenResponse.data);
```
JamesTLopez, line 76: Minor, can we add a line break before and after the condition for more visual clarity

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -113,33 +126,44 @@ export const createCollaborators = async ({
...
+
+		if (!applicationResult.success) {
+			return applicationResult;
+		}
+
+		const application = applicationResult.data;
+
+		// TODO: Valid states for actions should be handled through the appStateManager
+		if (!(application.state === 'DRAFT')) {
+			return failure('INVALID_STATE_TRANSITION', `Can only add Collaborators when Application is in state DRAFT`);
```
demariadaniel, line 151: Yeah there are a number of similar conditional checks in the app like this that could be moved into state manager functions. 
In this case though there's no state transition so do you think stateManager is necessary?

joneubank: this action is only valid when the application is in some states, so the action should be added to the state manager. doesn't need a state transition, just a check if the action can be pefformed

apps/ui/src/providers/Providers.tsx
joneubank: oops, nice catch

apps/api/src/controllers/stateManager.ts
```diff
@@ -84,49 +86,71 @@ export class ApplicationStateManager extends StateMachine<ApplicationStateValues
 	private _application: ApplicationRecord;
 	public readonly initState: ApplicationStateValues;
 
-	_canPerformAction(action: ApplicationStateEvents) {
+	_stateTransitionFailure(action: ApplicationStateEvents) {
+		return failure(
+			'INVALID_STATE_TRANSITION',
+			`Cannot perform action "${action}" on application with state "${this.getState()}"`,
+		);
+	}
```
demariadaniel, line 94: Good addition 👍

apps/api/src/routes/signatureRouter.ts
```diff
@@ -170,41 +218,98 @@ signatureRouter.post(
...
-				const result = await deleteApplicationSignature({
-					applicationId: Number(applicationId),
-					signee: signee,
-				});
+					// Validate Query Params
+					const queryValidationResult = deleteSignatureQuerySchema.safeParse(request.query);
+					if (!queryValidationResult.success) {
+						response.status(400).send({
+							error: 'INVALID_REQUEST',
+							message: `5Signee parameter must be either 'APPLICANT' or 'INSTITUTIONAL_REP'.`,
```
surajgoraya, line 246: Typo:
```suggestion
							message: `Signee parameter must be either 'APPLICANT' or 'INSTITUTIONAL_REP'.`,
```

However, regarding even having a parameter, I was hoping that, maybe as a future refactor this parameter would no longer be required. We can technically dynamically determine if they're a applicant or institutional rep through auth, so I feel like having this parameter feels like repetition.

Can DAC members delete parts of a document? if not, then we should be good to remove this in the future and just use Auth.

surajgoraya (commented):
The big one! Changes are looking ok so far! :shipit:  

Focused most of my attention to the signature related functionality, will take a look again in-case I've missed things elsewhere.

apps/api/src/utils/buildQueryParams.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
demariadaniel, line 1: Copyright

apps/api/src/server.ts
```diff
@@ -28,23 +28,27 @@ import { getHealth, Status } from '@/app-health.js';
...
 import signatureRouter from './routes/signatureRouter.ts';
 import sessionMiddleware from './session/sessionMiddleware.js';
 
+const logger = BaseLogger.forModule('server');
+
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
-const API_DOCS_PATH = `api-docs`;
+const API_PATH_DOCS = `/api-docs`;
```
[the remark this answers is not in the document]
joneubank: Thats an idea worth considering. We have a lot of hardcoded paths though, for all of our api endpoints for instance. I see no reason to treat this path differently.

apps/api/src/service/fileService.ts
```diff
@@ -22,11 +22,13 @@ import fs from 'fs';
 
 import { type PostgresDb } from '@/db/index.js';
 import { files } from '@/db/schemas/files.js';
-import logger from '@/logger.ts';
+import BaseLogger from '@/logger.ts';
 import { failure, success, type AsyncResult } from '@/utils/results.js';
 import { FileType } from '@pcgl-daco/data-model';
 import { eq } from 'drizzle-orm';
-import { PostgresTransaction, type FilesModel, type JoinedApplicationRecord } from './types.ts';
+import { PostgresTransaction, type FilesModel, type FilesRecord, type JoinedApplicationRecord } from './types.ts';
```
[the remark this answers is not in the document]
joneubank: As much as possible I'm letting the software manage and organize the imports, and this simplification has no bearing on me or my coding or reading effort. I'm not sure if there is a community standard for this atm, but generally I like to keep each type imported as a type. If in the future i want to import a function from this source then I won't need to add a second import line or rewrite this back to having type in front of each imported type.

Thanks for highlighting this import block though, I have found some missed type declarations and reorganized this section properly.

apps/ui/src/api/mutations/useDeleteCollaborator.ts
```diff
@@ -23,12 +23,12 @@ import { fetch } from '@/global/FetchClient';
...
+import { type ListCollaboratorResponse } from '@pcgl-daco/data-model';
 import { withErrorResponseHandler } from '../apiUtils';
 
 const useDeleteCollaborator = () => {
 	return useMutation<
-		CollaboratorsResponse[],
+		ListCollaboratorResponse,
 		ServerError,
 		{ applicationId: number | string; collaboratorId: number | string }
 	>({
```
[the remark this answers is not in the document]
joneubank: I don't know why the build continued working with this... thats concerning. Nice catch.

apps/ui/src/api/useGetUser.ts
```diff
@@ -0,0 +1,56 @@
...
+
+				switch (response.status) {
+					case 404:
+						error.message = translate('errors.http.404.title');
+						error.errors = translate('errors.http.404.message');
+						break;
+				}
+
+				throw error;
+			}
```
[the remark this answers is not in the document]
joneubank: This file is a duplicate of the one moved into `/queries/useGetUser.ts`, and that one is updated correctly. This is likely a merge conflict. I will remove this file.

apps/api/src/resources/swagger.yaml
```diff
@@ -634,14 +634,14 @@ paths:
     delete:
```
demariadaniel, line 634: We should create a follow up ticket for adding Auth on Swagger?

apps/api/src/controllers/applicationController.ts
joneubank: nice catch. code needs to be used, this is incomplete.

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -41,7 +49,10 @@ export const createCollaborators = async ({
 	application_id: number;
 	user_id: string;
 	collaborators: CollaboratorDTO[];
-}) => {
+}): AsyncResult<
+	ListCollaboratorResponse,
+	'UNAUTHORIZED' | 'SYSTEM_ERROR' | 'NOT_FOUND' | 'INVALID_STATE_TRANSITION' | 'DUPLICATE_RECORD'
```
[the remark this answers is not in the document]
joneubank: Yes that can and should be done. Will need a followup PR.

apps/api/src/controllers/signatureController.ts
```diff
@@ -20,32 +20,37 @@
...
 
 /**
  * Gets a signature for an application.
  * @param application_id - The Application ID
  * @returns Success with the signature and signed at time, properties may be `null` if not yet signed / Failure with Error.
  */
-export const getApplicationSignature = async ({ applicationId }: { applicationId: number }) => {
-	if (!isPositiveInteger(applicationId)) {
-		return failure('InvalidParams', 'Error: Application ID MUST be a positive number greater than or equal to 1.');
-	}
```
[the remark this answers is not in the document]
joneubank: Yes, the function signature already requires the `applicationId` variable to be a number so we're free to use it here without checking. This function also handles "not found" applications (which will happen with non positive ints or floats) so we're covered. The routers can perform the isPositiveInteger check when converting to a number just to provide clearer user feedback about invalid request parameters.

apps/api/src/controllers/signatureController.ts
```diff
@@ -54,31 +59,43 @@ export const getApplicationSignature = async ({ applicationId }: { applicationId
...
-	return result;
+	const updatedSignature =
+		signee === 'APPLICANT' ? result.data.applicant_signature : result.data.institutional_rep_signature;
+	if (updatedSignature) {
+		return success({
+			id: result.data.application_id,
+			signedAt: (signee === 'APPLICANT'
+				? result.data.applicant_signed_at
+				: result.data.institutional_rep_signed_at
+			)?.toISOString(),
```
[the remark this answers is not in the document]
joneubank: I am being explicit here but agree its not strictly required for the reason you stated... express wil stringify this correctly. I validated this with some testing in node shows that the default JSON.stringify behaviour for dates (ie. `Date.toString()` ) uses the `toISOString()`.

However, since I typed the return objet to use the DTO, and the DTO states this will be a string, I need to have it typed correctly to satisfy the type checker. So... now we're forced to set the output formatting explicitly, which is probably best.

apps/api/src/routes/signatureRouter.ts
joneubank: Can be updated in future work. I was trying not to modify the interfaces here, so I didn't need to change the UI interactions with the API... that would make this PR much much too big, rather than just too big.

apps/api/src/utils/buildQueryParams.ts
joneubank: Didn't mean to commit this file. Its not used. Removing.

apps/api/src/resources/swagger.yaml
joneubank: Oh interesting, good point. This should likely interact with your logged in session, but needs further investigation.

demariadaniel: Discussed offline - Auth gets applied on Swagger, but userId fields are not functional.
Implementing auth/swagger should be a follow up ticket.

demariadaniel (approved):
Tested and working locally and feedback's been addressed
Ready to merge some major changes 🚀

apps/api/src/routes/applicationRouter.ts
```diff
@@ -21,336 +21,506 @@ import {
...
+						return;
+					}
+					case 'NOT_FOUND': {
+						response.status(404).json({ error: result.error, message: result.message });
+						return;
+					}
 				}
-				response.send({ message: result.message, errors: String(result.errors) });
+			} catch (error) {
+				response.status(500).json({ error: 'SYSTEM_ERROR', message: 'Unexpected error.' });
```
Azher2Ali, line 187: Should we consider Adding logs for unexpected errors can help with debugging?
`console.error('Unexpected error in editApplication:', error);`

apps/api/src/routes/applicationRouter.ts
```diff
@@ -21,336 +21,506 @@ import {
...
+applicationRouter.post(
+	'/approve',
+	authMiddleware({ requiredRoles: ['DAC_MEMBER'] }),
+	async (
+		request,
+		response: ResponseWithData<{ message: string; data: ApplicationRecord }, ['INVALID_REQUEST', 'SYSTEM_ERROR']>,
+	) => {
+		const { applicationId }: { applicationId?: number } = request.body;
+
+		if (typeof applicationId !== 'number' || !applicationId) {
```
Azher2Ali, line 359: Can we replace this with if `(!Number.isInteger(applicationId)) {` 

`!Number.isInteger(applicationId)` ensures `applicationId `is a valid integer.

apps/api/src/routes/applicationRouter.ts
```diff
@@ -21,336 +21,506 @@ import {
...
-			}
+			return;
+		}
+
+		try {
+			const result = await dacRejectApplication({ applicationId });
 
-			res.status(status).send({ message, errors });
+			if (result.success) {
+				response.status(200).send({
```
Azher2Ali, line 421: `.json()` explicitly returns JSON, making responses predictable.

apps/api/src/routes/applicationRouter.ts
```diff
@@ -21,336 +21,506 @@ import {
...
+	async (
+		request,
+		response: ResponseWithData<
+			{ message: string; data: ApplicationRecord },
+			['INVALID_REQUEST', 'SYSTEM_ERROR', 'UNAUTHORIZED']
+		>,
+	) => {
+		const { applicationId } = request.body;
+
+		if (!applicationId || !isPositiveInteger(applicationId)) {
```
Azher2Ali, line 409: Can we consider Removing redundant !applicationId check and kept only isPositiveInteger(applicationId) for validation.
```
if (!isPositiveInteger(applicationId)) {
			response.status(400).json({
```

apps/api/src/routes/applicationRouter.ts
joneubank: Logging is good to consider. In this case, no new information is surfaced here and no logical decisions other than to format the response. The logging should take place when the unexpected outcome occurs (and we are logging in those cases, usually the catch block of the controller function before we return the `Failure<'SYSTEM_ERROR'>`. We also have logging for all web requests and their responses so we'll see the 500 occur.

apps/api/src/routes/applicationRouter.ts
joneubank: We should probably use the body validation function to validate the applicationId using a zod schema.

apps/api/src/routes/applicationRouter.ts
joneubank: Great catch, updating many cases across the app.

Issue #115, opened 2024-12-18
Endpoint to Submit Application


Buwujiu:
## Descriptions
Build a REST endpoint to let user submit application. The endpoint should allow:
- Applicant and Rep to submit application.
- Applicant and Rep can only submit when their signature is provided.
      -  mock the signature for now since we do not have a signature lib yet.
## Acceptance Criteria
1. If applicant submits, app status should become Rep Review.
2. if Rep submits, app status should become DAC Review.
3. Design the endpoint interface, define the query params, for example, could accept an app id.
4. Clearly document how to use the endpoint in swagger, add notes to each query params, add sample payload ect..
5. As a result of successful submission, should return a submission id, new app status, timestamps.

Buwujiu:
Has a dependency on the state transition, will wait for Dan's PR to merge first.

Buwujiu:
Azher is working on PR feedbacks.

Buwujiu:
Will split this to two endpoints, Azher is working on the PR feedbacks

Buwujiu:
Tested in the API integration ticket, looks good!

PR #224, opened 2025-02-18
#115 : Add endpoint to submit application with signature validation

Azher2Ali:
#115 : Add endpoint to submit application with signature validation

## Summary

This PR introduces a new REST API endpoint that allows an Applicant or Representative (Rep) to submit an application. The submission requires a signature (mocked for now).

If an Applicant submits, the application's status transitions to Rep Review.
If a Rep submits, the application's status transitions to DAC Review.
The response includes a submission ID, the new application status, and timestamps. 

## Description of Changes

**API Service**

- Added new endpoint:

`POST /applications/{appId}/submit`

**Query Parameters:**

- appId (path param): The unique identifier of the application to be submitted.
- Request Body:

```
{
  "submittedBy": "applicant", // or "rep"
  "signature": "mock-signature-value"
}
```

- Response:

```
{
  "submissionId": "12",
  "newStatus": "Rep Review",
  "submittedBy": "applicant",
  "timestamp": "2025-02-18T12:00:00Z"
}
```
 **Rules Implemented:**

- An Applicant can submit only when their signature is provided → Status becomes Rep Review.
- A Rep can submit only when their signature is provided → Status becomes DAC Review.
- Mocked signature validation for now (until a signature library is integrated).
- Error handling for invalid users, missing signatures, or invalid application states.

### Special Instructions
```
pnpm build:all
```

## Readiness Checklist

- [ ] **Self Review**
- [ ] **Labels Added**
- [ ] **Local Testing**
unit test case


surajgoraya:
This is for the "Sign and Submit" screen correct? If so, the signature sent in is going to be a base64 encoded image. But I'm also tracking the "createdAt" timestamp on the front end which is created when the submitter signs and saves the signature, this is similar to what OHCRN's doing. Do we want to only track this in the backend?

Azher2Ali:
> This is for the "Sign and Submit" screen correct? If so, the signature sent in is going to be a base64 encoded image. But I'm also tracking the "createdAt" timestamp on the front end which is created when the submitter signs and saves the signature, this is similar to what OHCRN's doing. Do we want to only track this in the backend?

Since we don't have the signature lib integerated yet, currently I have just added a mock signature. Once the validate signature method is implemented, we can further integerate it.

apps/api/src/controllers/applicationController.ts
```diff
@@ -210,3 +210,44 @@ export const rejectApplication = async ({ applicationId }: { applicationId: numb
...
+		// Transition application to the next state (e.g., under review)
+		const submissionResult = await appStateManager.submitDraft();
+
+		if (!submissionResult.success) {
+			return failure(submissionResult.message || 'Failed to submit application.', 'StateTransitionError');
+		}
+
+		// Save the application with its new state
+		const update = { state: appStateManager.state, submitted_at: new Date() };
+		const updatedResult = await service.findOneAndUpdate({ id: applicationId, update });
```
demariadaniel, line 244: `appStateManager.submitDraft()` already performs the application record update in the `_dispatchAndUpdateAction` function so these lines can be removed

apps/api/src/routes/application-router.ts
```diff
@@ -295,4 +296,66 @@ applicationRouter.post('/applications/reject', jsonParser, async (req, res) => {
 		});
 	}
 });
+
+applicationRouter.post('/applications/submit', jsonParser, async (req, res) => {
+	const { applicationId, role, signature } = req.body;
+
+	if (!applicationId || !role || !signature) {
+		res.status(400).json({ message: 'Missing required fields.' });
+	}
```
demariadaniel, line 305: Should replace this with use of `withSchemaValidation`

apps/api/src/routes/application-router.ts
```diff
@@ -295,4 +296,66 @@ applicationRouter.post('/applications/reject', jsonParser, async (req, res) => {
...
+
+	if (!applicationId || isNaN(parseInt(applicationId))) {
```
demariadaniel, line 307: `!applicationId` is already checked in the above conditional
Also we can replace `isNaN(parseInt(applicationId)` with `isPositiveNumber` from `/utils/routes`

apps/api/src/routes/application-router.ts
```diff
@@ -295,4 +296,66 @@ applicationRouter.post('/applications/reject', jsonParser, async (req, res) => {
...
+	if (signature !== 'mock_signature') {
+		res.status(400).json({ message: 'Invalid signature.' });
+	}
+
+	let newStatus = '';
+	if (role === 'APPLICANT') {
+		newStatus = 'Rep Review';
+	} else if (role === 'REP') {
+		newStatus = 'DAC Review';
+	} else {
```
demariadaniel, line 324: This `newStatus` variable is unused, this block should be changed to `if (!(role === 'APPLICANT' || role === 'REP'))`

demariadaniel (changes requested):
- Needs merge w/ main, some conflicts to address
- Needs unit test for the controller function
- Left feedback, will get back to you ASAP re: submit requirements

apps/api/src/controllers/applicationController.ts
```diff
@@ -210,3 +210,44 @@ export const rejectApplication = async ({ applicationId }: { applicationId: numb
 		return failure(message, error);
 	}
 };
+
+export const submitApplication = async ({ applicationId }: { applicationId: number }) => {
```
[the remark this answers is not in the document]
demariadaniel: OK so I sorted this out.

line 231 should include a check for `if (appStateManager.state !== ApplicationStates.DRAFT || appStateManager.state !== ApplicationStates.INSTITUTIONAL_REP_REVIEW)`

line 236 depending on Application state either call `await appStateManager.submitDraft();` or `await appStateManager.approveRepReview();`

`approveRepReview` can be renamed `submitRepReview`

packages/validation/src/routes/applicationRoutes.ts
```diff
@@ -62,3 +62,11 @@ export const editApplicationRequestSchema = z.object({
 		params: { violation: 'noEmptyObject' },
 	}),
 });
+
+export const submitApplicationRequestSchema = z
+	.object({
+		applicationId: z.string().min(1),
+		role: z.enum(['APPLICANT', 'REP']),
+		signature: z.string().min(1),
```
surajgoraya, line 70: The signature is confirmed to always be a base64 Image so adding the same validation we have in the `signatureRoutes.ts` file in the validation package makes sense here. Also, for role (called `signee` in the sig API), lets make it consistent and use INSTITUTIONAL_REP. 

Also given this is being passed through the body, appID will be a number and can be validated (unlike in the params validation).

```suggestion
	.object({
		applicationId: z.number().nonnegative(),
	      role: z.literal('APPLICANT').or(z.literal('INSTITUTIONAL_REP')),
	      signature: z.string().regex(BASE64_IMAGE),
```

apps/api/src/routes/applicationRouter.ts
```diff
@@ -338,4 +348,64 @@ applicationRouter.post(
...
+
+					if (errors === 'ApplicationNotFound' || errors === 'Application record is undefined') {
+						status = 404;
+						message = 'Application not found.';
+					} else if (errors === 'RejectionConflict') {
+						status = 409;
+						message = 'Rejection conflict detected.';
+					} else if (errors === 'InvalidState') {
+						status = 400;
+						message = 'Invalid application state.';
```
surajgoraya, line 397: This looks like a good place to do a switch case, but not required!

apps/api/src/resources/swagger.yaml
```diff
@@ -537,7 +537,43 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ApplicationResponse'
+        '400':
+          description: Invalid request or application cannot be submitted
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ClientErrors'
```
surajgoraya, line 564: Since we have the request validation middleware working on this route now you should also list `$ref: '#/components/responses/RequestValidationError'` as a possible returnable error

surajgoraya (commented):
A few small things! Looking good overall!

apps/api/src/resources/swagger.yaml
Azher2Ali: Didn't notice this was merged to `main `

packages/validation/src/routes/applicationRoutes.ts
Azher2Ali: Great suggestion! Updated the code now..

joneubank (changes requested):
I have some requests to restructure how this is built. Importantly, I think we need to clearly separate our actions into their own endpoints. An applicant and institutional rep are performing separate actions, and these don't need to be combined into a single endpoint.

1. Lets separate this into two endpoints: `/application/{id}/submit` for the application submission, and `/application/{id}/rep/approve` for the institutional rep review.
2. Remove from the API body the user role. We will validate this from user session.
3. Remove from the API body the signature data, this is already saved into the application through the /edit/signature endpoint.

apps/api/src/routes/applicationRouter.ts
```diff
@@ -408,6 +463,88 @@ applicationRouter.post(
...
+		res.status(200).json(updatedApplication);
+	}),
+));
+
+// Endpoint for reps to request revisions
+applicationRouter.post(
+	'/rep/request-revisions',
+	jsonParser,
+	withBodySchemaValidation(applicationRevisionRequestSchema, apiZodErrorMapping, async (req, res) => {
+		const { applicationId, revisionData, role } = req.body;
```
joneubank, line 514: Why did we move the application ID to the request body? It can still be used in the path, for consistency.

apps/api/src/routes/applicationRouter.ts
```diff
@@ -408,6 +463,88 @@ applicationRouter.post(
...
 			}
 		},
-	),
+
+// Endpoint for reps to request revisions
+applicationRouter.post(
+	'/dac/request-revisions',
+	jsonParser,
+	withBodySchemaValidation(applicationRevisionRequestSchema, apiZodErrorMapping, async (req, res) => {
+		const { applicationId, revisionData, role } = req.body;
```
joneubank, line 472: Why did we move the applicationId from the path into the body? The feedback before had examples like:

Applicant submit: `application/{id}/submit`
DAC request revision: `application/{id}/dac/request-revision`
Rep request revision: `application/{id}/rep/request-revision`

Its consistent with existing endpoints, and is consistent with the REST pattern.

apps/api/src/controllers/applicationController.ts
```diff
@@ -257,6 +257,102 @@ export const submitRevision = async ({ applicationId }: { applicationId: number
 	}
 };
 
+export const requestApplicationRevisionsByDac = async ({
+	applicationId,
+	role,
```
joneubank, line 262: This function probably does not need the `role` input parameter. This will only get called when the system intends to take the DAC revisions action, all the role based approval should happen before this is called. In fact, we do perform this role check in the router.

Strictly speaking this can be left in, but we should standardize how we do permission checking instead of passing partial identity information as parameters into our functions. I'll approve this for merging but will likely refactor it along with the other API auth changes.

joneubank (changes requested):
Pretty much ready to go, I'd just like to change the endpoints themselves to use the RESTful URI pattern, example: `application/{application_id}/dac/request-revisions`

joneubank (approved):
Something has changed between my review and what is now the diff with develop. The changes I requested were on code that was marked as added, but when I look now they are shown as already merged with develop. I don't know what the confusion is but the changes I requested are no longer part of this PR. There may be some other refactoring necessary afterwards but I can approve this PR.

apps/api/src/routes/applicationRouter.ts
Azher2Ali: It was added in the body, so that it can be fetched along with revision data

demariadaniel (approved):

PR #217, opened 2025-02-13
#116 & #118 : Add endpoints for Rep and DAC to request application revisions

Azher2Ali:
#116 & #118: Add endpoints for Rep and DAC to request application revisions

## Summary

Added REST endpoints to allow Rep and DAC to request revisions on an application. These endpoints ensure that only requested sections can be modified and that the application status updates accordingly.

## Description of Changes
**API Changes**
Added a **POST** `/applications/{appId}/rep-revisions` endpoint to allow Reps to request revisions.

- Updates application status from "Rep Review" to "Rep Revisions".
- Allows requesting revisions on specific sections:
- Applicant Information
- Institutional Rep
- Collaborators
- Project Info
- Requested Study
- Ethics
- Supports adding a general comment.
- Ensures that only one application can have requested revisions at a time.
- Includes input validation and error handling.

Added a **POST** `/applications/{appId}/dac-revisions` endpoint to allow DAC to request revisions.

- Updates application status from "DAC Review" to "DAC Revisions".
- Allows requesting revisions on the same sections as the Rep.
- Supports adding a general comment.
- Ensures only one application can have requested revisions at a time.
- Includes clear success and error messages.

**Error Handling & Validation**

- Ensures revisions can only be requested for allowed sections.
- Ensures the application is in the correct status before requesting revisions.
- Returns appropriate HTTP status codes (e.g., 400 Bad Request for invalid inputs, 404 Not Found if the application does not exist).

### Special Instructions
Before running these changes, you will need to install `package name`:
```
pnpm build:all
```
## Readiness Checklist

- [x] **Self Review**
  - I have performed a self review of code
  - I have run the application locally and manually tested the feature
- [x] **Labels Added**
  - Label is added for each package/app that is modified (`api`, `ui`, `data-model`, etc.)
  - Label is added for the type of work done in this PR (`feature`, `fix`, `chore`, `documentation`)
- [x] **Local Testing**
  - Successfully built all packages locally
  - Successfully ran all test suites, all unit and integration tests pass
- [x] **Updated Tests**
  - Unit and integration tests needs to be added
  
Unit tests

Issue #118, opened 2024-12-18
Endpoint to Allow DAC Request Revisions


Buwujiu:
## Descriptions
Build a REST endpoint to allow DAC request revisions. 

Refer to mockups:
https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL%2FDesign%2FMockups?node-id=2488-18439&t=eWmr8NiU3lVp8zzm-4

## Acceptance Criteria
- Design the interface, define inputs and outputs.
- Write error handling
- The application status will change from DAC Review to DAC Revision after revisions are requested. 
- Should give clear message on success request. Clear message if request fails.
- DAC can only request revisions on one app at a time.
- DAC can request revisions on any of these sessions: applicant information, institutional rep, collaborators, project info, requested study, ethics. They can also leave a general comment. See mock:

![Image](https://github.com/user-attachments/assets/0f2403e6-7b4d-4892-b408-c7d5a22a98da)

Buwujiu:
Will test in dev using UI. Closing ticket!

Issue #116, opened 2024-12-18
Endpoint to Allow Rep Request Revisions


Buwujiu:
## Descriptions
Build a REST endpoint to allow rep to request revisions on an application. 

## Acceptance Criteria
- Design the interface, define inputs and outputs.
- Write error handling 
- When rep requests revisions, the app status will change from Rep Review to Rep Revisions
- Rep can only request revisions on one app at a time
- Add Unit test
- Rep can request revisions on any of these sessions: applicant information, institutional rep, collaborators, project info, requested study, ethics. They can also leave a general comment. See mock:

![Image](https://github.com/user-attachments/assets/89e53dfe-8418-4c12-8142-87c9057cc4f1)

edsu7:
@Buwujiu is this endpoint hooked up on UI? If not I'm unable to test through swagger API due to authorization.

PR #238, opened 2025-02-26
Chore: Revisions to PR #217

demariadaniel:
## Summary

[Revisions from PR discussion](https://github.com/Pan-Canadian-Genome-Library/daco/pull/217/)
    - Correct type usage 
    - Remove 'comments' field
    - Return revision record?
    - Revision Service?
    - Zod Request Validation

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/116
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/118

## Description of Changes
- Remove 'comments' field: `comments` is a single field on the revisionRequest, schema and in the UI mockups it shares the same modal as the other fields. So I think it makes sense to group `comments` with all the `reviewData` in req.body rather than having them separate. I don't think we need `comments || comments_str`

- Change `reviewData` to `revisionData`
- Flatten Request Body object

- Correct type usage: `service.createRevisionRequest` was using type `any` which we try to avoid in our projects unless totally necessary. I updated the Types for ReviewData usage and added an explicit Return type to the Service method.
Note that in `service.createRevisionRequest` this means you don't have to do any camelCase -> snake_case aliasing because the object passed in is using the same keys as the database expects.
I use the names `Model` to denote insert/update types and `Record` to signify records retrieved from the database.
    
- Zod Request Validation: We should be adding a Zod schema validator to this endpoint. I haven't added one here yet. See application-router L74 `withSchemaValidation(editApplicationRequestSchema...)` for an example.
The Zod validator should use camelCase (because that's what the UI uses) then in applicationController we can add a function to map it to snake case.

- `role &&`: application-router has a conditional check for user `role` which was using `&&`, this would fail the check `!role && (role !== 'REP' || role !== 'DAC')` because `role='ANONYMOUS'` is truthy so it would pass

- Return revision record: I think we discussed returning an Application record from this endpoint which is what you had in place. I changed it to use the `getApplicationWithContents` function so all details are returned.

- Revision Request Service: We don't need to do this as part of this PR, just bringing it up for discussion. 
IIRC we discussed at planning that we could just add this method to the applications service. If we need more Revision Request functionality we will want to build a full Revision Request service. This should be a separate ticket.

## Readiness Checklist

- [ ] **Self Review**
- [ ] **PR Format**
- [ ] **Labels Added**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation

### DIFF
--- a/apps/api/src/controllers/applicationController.ts
+++ b/apps/api/src/controllers/applicationController.ts
@@ -25,10 +25,10 @@ import { type ApplicationListRequest } from '@/routes/types.js';
 import { applicationSvc } from '@/service/applicationService.js';
 import {
 	type ApplicationContentUpdates,
-	type ApplicationModel,
 	type ApplicationRecord,
 	type ApplicationService,
-	type ReviewApplication,
+	type RevisionRequestModel,
+	type RevisionRequestRecord,
 } from '@/service/types.js';
 import { failure, success, type AsyncResult } from '@/utils/results.js';
 import { aliasApplicationRecord } from '@/utils/routes.js';
@@ -221,13 +221,11 @@ export const requestApplicationRevisions = async ({
 	applicationId,
 	role,
 	reviewData,
-	comments,
 }: {
 	applicationId: number;
 	role: string;
-	reviewData: ReviewApplication;
-	comments?: string;
-}): AsyncResult<ApplicationModel> => {
+	reviewData: RevisionRequestModel;
+}): AsyncResult<RevisionRequestRecord> => {
 	try {
 		const database = getDbInstance();
 		const service: ApplicationService = applicationSvc(database);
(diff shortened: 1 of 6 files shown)


apps/api/src/controllers/applicationController.ts
```diff
@@ -254,9 +252,13 @@ export const requestApplicationRevisions = async ({
 			return failure(revisionResult.message || 'Failed to reject application.', 'StateTransitionError');
 		}
 
-		service.createRevisionRequest({ applicationId, reviewData, comments_str: comments });
+		const revisionRequestResult = await service.createRevisionRequest({ applicationId, revisionData });
```
demariadaniel, line 255: Need to handle potential failure here

apps/api/src/resources/swagger.yaml
```diff
@@ -685,80 +687,62 @@ components:
         role:
           type: string
           description: 'Role of the reviewer'
-          example: DAC
-        comments:
-          type: string
-          description: 'additional comments'
-        reviewData:
+          example: DAC_MEMBER
```
demariadaniel, line 690: Jon's started defining user roles in his auth PRs so I updated these values

JamesTLopez (approved):
lgtm, simple changes 👍

apps/api/src/controllers/applicationController.ts
Azher2Ali: Failure cases are already being handled in L257 - https://github.com/Pan-Canadian-Genome-Library/daco/pull/238/files/3157eccd2d8fc55310197d0c1a6614b9048ca0f9#diff-9ce0914fc976635505ddd7a5058c3ee26cc63dfa9b78d43b23e2cb9e9a0ce8e0R257

Issue #168, opened 2025-01-22
Endpoint To Delete Collaborators


Buwujiu:
## Descriptions
Write a REST endpoint to delete collaborators from an application.

## Acceptance Criteria
- Only applicant can delete collaborators from their application.
- Delete one collaborator at a time.
- Can only delete collaborators when app is in Draft.
- Proper error handling for server errors and application not found.

edsu7:
|Testing the following| result|
|--|--|
|Deleting collaborator from application| Success |
|Block Deleting collaborator in different application| Failure |
|Errors out successfully when deleting non-existent collaborator | Success|

Notes:
- collaborator belongs to application `36` and not `38` but specifying `collaborator id` is enough to delete
![Image](https://github.com/user-attachments/assets/bb0b1059-9d36-4365-b04f-b0c04be7c487)

Buwujiu:
Thanks @edsu7 ! closing ticket!

PR #241, opened 2025-02-26
#168 ✂️ Delete Collaborators endpoint

demariadaniel:
## Summary

Endpoint to Delete Collaborator records

### Related Issues

- [168](https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=94807223&issue=Pan-Canadian-Genome-Library%7Cdaco%7C168)

## Description of Changes
- Service/Controller/API/Swagger/Validation additions for Delete Collaborators function
- `request-utils/withSchemaValidation` type updates to insure req.body is properly typed
- Minor docs correction

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -98,3 +98,48 @@ export const createCollaborators = async ({
...
+	if (!applicationResult.success) {
+		return applicationResult;
+	}
+
+	const application = applicationResult.data;
+
+	// TODO: Add Real Auth
+	// Validate User is Applicant
+	if (!(user_id === application.user_id)) {
+		return failure('Unauthorized, cannot create Collaborators', 'Unauthorized');
```
demariadaniel, line 133: Last Todo: Auth

JamesTLopez (commented):
Minor stuff, functionally lgtm

apps/api/src/resources/swagger.yaml
```diff
@@ -394,6 +394,40 @@ paths:
...
+              type: object
+              $ref: '#/components/requests/DeleteCollaborators'
+      responses:
+        '500':
+          description: 'Error deleting collaborators'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ServerErrors'
+        '400':
```
surajgoraya, line 417: Small but with the most recent commit I added a definition for our Zod Validation error, it might be good to reference here as a possible return since you're using `withSchemaValidaiton`, 

```yaml
$ref: '#/components/responses/RequestValidationError'
```

apps/api/src/service/collaboratorsService.ts
```diff
@@ -84,6 +84,24 @@ const collaboratorsSvc = (db: PostgresDb) => ({
...
+		try {
+			const deletedRecord = await db.delete(collaborators).where(eq(collaborators.id, id)).returning();
+
+			if (!deletedRecord[0]) {
+				throw new Error(`Error deleting collaborator with ${id}, no record deleted`);
+			}
+
+			return success(deletedRecord[0]);
+		} catch (err) {
+			const message = `Error at deleteCollaborators`;
```
surajgoraya, line 100: Small but might be good to try and log the ID of the collaborator being deleted in the catch statement

surajgoraya (commented):
Super minor stuff from my end! Overall looks pretty good to me!

apps/api/src/routes/collaboratorsRouter.ts
```diff
@@ -63,4 +63,39 @@ collaboratorsRouter.post(
 	}),
 );
 
+/**
+ * Delete Collaborator
+ */
+collaboratorsRouter.post(
+	'/collaborators/delete',
```
surajgoraya, line 70: I think we should remove repeating the `/collaborators/` string prepending the routes and use a similar pattern to the auth / signature router by attaching it to where we define it in `server.ts`. 

I've made this update for the Applications router as well in #256

surajgoraya: Also I wonder if we should use the `DELETE` HTTP request here instead of POST? Not that it makes a huge difference, just wondering!

apps/api/src/routes/collaboratorsRouter.ts
```diff
@@ -18,12 +18,12 @@
  */
 
 import bodyParser from 'body-parser';
-import express from 'express';
+import express, { Request, Response } from 'express';
```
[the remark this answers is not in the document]
demariadaniel: This was related to the type conflicts I was having in request utils but it maybe isn't needed any more, taking a look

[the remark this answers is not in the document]
demariadaniel: OK was able to remove this and use implicit types. `type` import was partly omitted b/c Req/Res are interfaces (but we were only using the `type` so fair callout).

apps/api/src/resources/swagger.yaml
demariadaniel: Nice thank you updated

apps/api/src/routes/collaboratorsRouter.ts
demariadaniel: Delete is a good suggestion, I will make the update

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -98,3 +98,48 @@ export const createCollaborators = async ({
...
+	}
+
+	if (!(application.state === 'DRAFT')) {
+		return failure(`Can only add Collaborators when Application is in state DRAFT`, 'InvalidState');
+	}
```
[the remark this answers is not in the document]
demariadaniel: In this case the ticket specifies that Collaborators can only be deleted in DRAFT
https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=94807223&issue=Pan-Canadian-Genome-Library%7Cdaco%7C168

packages/request-utils/src/validation.ts
```diff
@@ -54,9 +53,9 @@ import { RequestValidationErrorResponse } from './responses.js';
 function withBodySchemaValidation<ReqBody>(
 	bodySchema: ZodSchema<ReqBody>,
 	zodErrorMapping: ZodErrorMap | undefined,
-	handler: RequestHandler<ParamsDictionary, any, any, qs.ParsedQs>,
-): RequestHandler {
-	return async (request: Request, response: Response, next: NextFunction) => {
+	handler: RequestHandler<ParamsDictionary, any, ReqBody, qs.ParsedQs>,
+): RequestHandler<ParamsDictionary, any, ReqBody, qs.ParsedQs> {
+	return async (request, response, next: NextFunction) => {
```
demariadaniel, line 58: I made these updates b/c I was not getting the typed `req.body` as expected

packages/validation/src/utils/functions.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Helper function to determine if value is a valid number and is positive.
+ * @param num Any integer value
+ * @returns True if the value is a valid number and is positive, false otherwise.
+ */
+export const isPositiveNumber = (num: number) => {
+	if (Number.isNaN(num) === false && num >= 0) {
+		return true;
+	}
+	return false;
+};
```
ciaranschutte, line 30: TS `number` covers floats too https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#the-primitives-string-number-and-boolean

I think this is technically correct because of how JS implements IEEE standards with 0 being positive by "default". In practical terms and based on this being a boolean check, I imagine it's use case may result in 0 causing unexpected results because it's a falsy js value.

apps/api/src/routes/collaboratorsRouter.ts
```diff
@@ -112,4 +112,46 @@ collaboratorsRouter.get(
...
 );
 
+/**
+ * Delete Collaborator
+ */
+collaboratorsRouter.delete(
+	'/:applicationId/:collaboratorId',
+	jsonParser,
+	withParamsSchemaValidation(collaboratorsDeleteParamsSchema, apiZodErrorMapping, async (request, response) => {
+		const { applicationId, collaboratorId } = request.params;
```
justincorrigible, line 122: 👍

packages/validation/src/utils/functions.ts
demariadaniel: I see what you're saying
Essentially this should be changed to `if (Number.isNaN(num) === false && Number.isInterger(num) && num > 0) {`

packages/validation/src/schemas.ts
```diff
@@ -62,13 +63,35 @@ export const collaboratorsSchema = z.object({
...
+	.object({
+		applicationId: z
+			.string()
+			.refine((id) => isPositiveNumber(parseInt(id)), { message: 'applicationId must be a positive number' }),
+		collaboratorId: z
+			.string()
+			.refine((id) => isPositiveNumber(parseInt(id)), { message: 'collaboratorId must be a positive number' }),
+	})
+	.required();
+
```
surajgoraya, line 94: Sorry didn't catch this earlier, It would probably be a good idea to place this within the `/routes/` directory in the package, likely within `src/routes/collaboratorsRoutes.ts`. 

This was a suggestion from Jon back when we started adding JSON body validation to the `/application` routes, and I think it helps reduce the length of this file. 

This file should really be refactored further whenever we get the chance.

packages/validation/src/utils/functions.ts
demariadaniel: Updated usage of isPositiveNumber & parseInt: [Update usage of isPositiveNum function](https://github.com/Pan-Canadian-Genome-Library/daco/pull/241/commits/a75d415956b355ad15f56814b5fc0291f495de97)

packages/validation/src/schemas.ts
demariadaniel: Thanks for the catch, moved these into their own file

apps/api/src/resources/swagger.yaml
```diff
@@ -352,6 +352,44 @@ paths:
               schema:
                 $ref: '#/components/responses/ApplicationResponse'
 
+  /applications/{applicationId}/submit-revision:
```
demariadaniel, line 355: Moved

surajgoraya (approved):
Changes LGTM! :shipit:

Issue #121, opened 2024-12-18
Endpoint - Applicant Closes Application


Buwujiu:
## Descriptions
Build a REST endpoint to allow applicant close application. 
This action can take place when the application is in Draft, or Rep Review, or DAC review state. 
Closed applications cannot be re-opened.

PR #245, opened 2025-02-28
#121: Add endpoint to allow applicants to close applications

Azher2Ali:
#121 - Add endpoint to allow applicants to close applications

### Summary:
Added a REST endpoint to allow applicants to close their application. The endpoint enforces role-based access control and ensures that only applications in specific states can be closed. Additionally, unit tests and error handling have been implemented to validate the functionality.

**Description of Changes:**

- New REST Endpoint: Created an endpoint that allows an applicant to close their application.
- State-Based Validation: Applications can only be closed if they are in Draft, Rep Review, or DAC Review states.
- Role-Based Access Control:
- Draft: Only the applicant can close.
- Rep Review: Only the applicant can close.
- DAC Review: The applicant and DAC members can close.
- Finality Rule: Once an application is closed, it cannot be reopened.
- Error Handling:
- Returns appropriate error responses for unauthorized attempts.
- Prevents closing an application in an invalid state.
- Unit Tests: Added test cases to validate the logic for different states and roles.

### Special Instructions
```
pnpm build:all
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
  - The PR title is properly formatted to match the pattern: `#121: Add endpoint to allow applicants to close applications `
- [x] **Labels Added**
  - Label is added for each package/app that is modified (`api`)
- [x] **Local Testing**
- [x] **Updated Tests**


apps/api/src/controllers/applicationController.ts
```diff
@@ -210,3 +210,73 @@ export const rejectApplication = async ({ applicationId }: { applicationId: numb
...
+			default:
+				return failure(`Cannot close application in state ${appStateManager.state}.`, 'InvalidState');
+		}
+
+		if (!closeResult.success) {
+			return failure(closeResult.message || 'Failed to close application.', 'StateTransitionError');
+		}
+
+		// Update the database with the new CLOSED state
+		const update = { state: ApplicationStates.CLOSED, updated_at: new Date() };
```
surajgoraya, line 272: Small suggestion, feel free to ignore it but some of our other Application functions tend to update the `updated_at` property within the service using SQL rather than the controller. Not a big deal if you don't change this but it might be a good idea!   

`findOneAndUpdate` actually already does the updated_at call in the service, so providing it here may be redundant since it gets overwritten

apps/api/src/routes/application-router.ts
```diff
@@ -295,4 +296,58 @@ applicationRouter.post('/applications/reject', jsonParser, async (req, res) => {
 		});
 	}
 });
+
+applicationRouter.post('/applications/:applicationId/close', jsonParser, async (req, res) => {
```
surajgoraya, line 300: It may be a good idea to update this to use our new `withBodySchemaValidation` middleware. You'll likely have to add a new Zod definition within the Validation package under the `/routes/applicationRoutes.ts` file.

apps/api/src/resources/swagger.yaml
```diff
@@ -349,6 +349,60 @@ paths:
...
+          required: true
+          description: The ID of the application to close.
+      responses:
+        '200':
+          description: Application successfully closed.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Applications'
+        '400':
```
surajgoraya, line 376: If you do add `withBodySchemaValidation`, you will also need to modify this to have our Zod Failure definition.

surajgoraya (changes requested):
Changes overall look good! Really good job with the tests, I do have small suggestions for the PR. Additionally, some changes have occurred with the Application Router so you'll likely need to merge and resolve conflicts.

I think adding `withBodySchemaValidation` would be a good idea for the post request body. Other suggestions are optional!

Also, the PR template is still missing the checklist, labels and details 📝

apps/api/src/controllers/applicationController.ts
```diff
@@ -252,3 +252,73 @@ export const submitRevision = async ({ applicationId }: { applicationId: number
...
+			return failure('Application is already closed.', 'StateConflict');
+		}
+
+		// Authorization logic based on state
+		const isApplicant = application.user_id === requesterId;
+		let closeResult;
+
+		switch (appStateManager.state) {
+			case ApplicationStates.DRAFT:
+				if (!isApplicant) {
```
demariadaniel, line 288: These `!isApplicant` checks don't need to be repeated, you could move this right after line 283 and only do it once

demariadaniel: Hmm I see it's a little more complicated than that.

What I see is that `!isApplicant` causes an error no matter which state.
Only Dac Review has an extra case where it could be a Dac Member.

So I think the switch case could be condensed to two conditionals instead of 4 different cases.
something like:
`
const canCloseDac = !isDacMember && appStateManager.state === ApplicationStates.DAC_REVIEW;
if (!isApplicant) { 
    if (!canCloseDac) { return failure(...) } // if you need Dac specific messaging
    return failure(...)
}
`

And then the `appStateManager.closeDraft();` etc can be handled in their own block

Just see a lot of repeat error messages which could be reduced

apps/api/src/controllers/applicationController.ts
```diff
@@ -252,3 +252,73 @@ export const submitRevision = async ({ applicationId }: { applicationId: number
...
+				return failure(`Cannot close application in state ${appStateManager.state}.`, 'InvalidState');
+		}
+
+		if (!closeResult.success) {
+			return failure(closeResult.message || 'Failed to close application.', 'StateTransitionError');
+		}
+
+		// Update the database with the new CLOSED state
+		const update = { state: ApplicationStates.CLOSED };
+		const updatedResult = await service.findOneAndUpdate({ id: applicationId, update });
```
demariadaniel, line 315: These are not needed, `appStateManager.closeDraft();` and other close functions already include an application update
See: https://github.com/Pan-Canadian-Genome-Library/daco/blob/main/apps/api/src/controllers/stateManager.ts#L119

apps/api/src/controllers/applicationController.ts
```diff
@@ -352,3 +352,67 @@ export const requestApplicationRevisionsByRep = async ({
...
+		if (appStateManager.state === ApplicationStates.CLOSED) {
+			return failure('Application is already closed.', 'StateConflict');
+		}
+
+		// Authorization logic based on state
+		const isApplicant = application.user_id === requesterId;
+
+		const canCloseApplication = isDacMember && appStateManager.state === ApplicationStates.DAC_REVIEW;
+
+		if (!isApplicant && !canCloseApplication) {
```
demariadaniel, line 387: This should be `||`, `if not applicant OR cannot close application`

demariadaniel: also `canCloseApplication` might be good to rename as `canCloseDacApplication` because it is actually pretty specific

demariadaniel: OK I reviewed the commit/discussion about this. What's missing here is the application state.

```
	if (!isApplicant && applicationState ===  DRAFT || applicationState ===  INSTITUTIONAL_REP_REVIEW ) {
	 return failure 
	 } else if ((!isApplicant || !canCloseDacApplication)) { 
  	return failure
	}
}
```

demariadaniel: edit: `if (!isApplicant || !canCloseApplication) { return failure ... }` covers all state use cases, so update `&&` to `||` here

demariadaniel (approved):
Looks good to me! :shipit:

surajgoraya (approved):
LGTM! PR Template is still blank, so just editing it real quick just for the sake of having a nice summary of the changes would be good, but otherwise this looks great!! 

Major functionality add :shipit:

Azher2Ali:
> LGTM! PR Template is still blank, so just editing it real quick just for the sake of having a nice summary of the changes would be good, but otherwise this looks great!!
> 
> Major functionality add :shipit:

Added description, totally missed it througout the feedback..

Issue #214, opened 2025-02-13
API to Get Signature By Application


Buwujiu:
## Descriptions
Build an API to get signature by application.
## Acceptance Criteria
- Only one signature is allowed for an application.

Buwujiu:
Tested in dev, working as expected!

PR #276, opened 2025-03-06
#214: API to Get Signatures by Application

surajgoraya:
## Summary

Adds an API to get signatures when provided an application ID. It also adds related tests, along with correcting minor spelling errors within the codebase.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/214

### Server
**✨ NEW:** 
- Added new endpoint `GET /signature` that gets both the Applicant and Signature Rep signatures (along with their respective signed at date times) for a given application.
   - Endpoint takes in a query param containing the application ID, and will return a 200 and a object on successfully finding an application with that ID
      - Noting that the object may contain null values if signatures haven't yet been signed, but an application does exist.  
   - Endpoint will return a 404 if an application is not found.
- Added a translation function to translate between DTO-case, and what our DB expects. 

**🐛 FIXES:**
- `/collaborators/:id` endpoint did not correctly pass Zod validation, letting incorrect values pass to the controller layer. This has been fixed.
- `/collaborators/:id` endpoint did not return a 404 when a application with a ID was not found, this has been fixed.
- Fixed some unrelated spelling issues within the codebase.

### Special Instructions
Before running these changes, you will need to rebuild the `@pcgl-daco/validation` and `@pcgl-daco/data-model` packages:

```bash
#in the root of the project folder:
pnpm i && pnpm build:all
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
  - image
- [x] Documentation


surajgoraya:
🤔  **Important Note:** Currently, this endpoint doesn't check ownership of the application when getting signatures. Since getting signatures will be available to both those who created the application (applicants) and institutional reps, we can't just limit it via a where clause in our DB call. 

I'm open to suggestions on how we can do this. Should I wait for Jons PR here?

Azher2Ali:
> 🤔 **Important Note:** Currently, this endpoint doesn't check ownership of the application when getting signatures. Since getting signatures will be available to both those who created the application (applicants) and institutional reps, we can't just limit it via a where clause in our DB call.
> 
> I'm open to suggestions on how we can do this. Should I wait for Jons PR here?

We can introduce Ownership Validation Based on User Roles
Since both applicants and institutional representatives should have access to signatures, we could introduce role-based access control (RBAC). 
If both applicants and institutional representatives should be able to sign an application, we need a way to verify that the user is either:

- The applicant who created the application.
- The institutional representative associated with the application.

1. Furthermore, by adding ownership checks before returning signatures `(GET /signature`).
2. Restricted signing permissions to the correct user `(POST /sign)`.
3. Ensured unauthorized users receive a 403 Forbidden instead of allowing unrestricted access.

But I can't really re-collect which PR who are reffering to which Jon is currently working, but if Jon's PR introduces changes related to ownership validation, it might be worth waiting to avoid redundant work. However, if his PR is unrelated, you could proceed with a simpler validation now and refine it later.

apps/api/src/controllers/signatureController.ts
```diff
@@ -20,20 +20,45 @@
...
+	const signatureRepo: SignatureService = signatureService(database);
+
+	const result = await signatureRepo.getApplicationSignature({ application_id: applicationId });
+
+	if (result.success) {
+		const aliasedResponse = aliasSignatureRecord(result.data);
+		return success(aliasedResponse);
+	}
+
+	return result;
```
JamesTLopez, line 47: This works, but its a little confusing. Imo, we should check if the `result.success` is FALSE and then return that result. Then we can have the aliasSignatureRecord step last.

```js
        const result = await signatureRepo.getApplicationSignature({ application_id: applicationId });

	if (!result.success) {
             return result;
	}

	const aliasedResponse = aliasSignatureRecord(result.data);
	return success(aliasedResponse);
	

```

JamesTLopez (approved):
🎸

Issue #230, opened 2025-02-20
Endpoint to Create File


Buwujiu:
## Descriptions
Create File - Insert new file entry, requires as inputs the file (content and filename), submitting user ID, and file type, returns new file object (including ID)

Issue #207, opened 2025-02-13
Ability to Upload Ethics File


Buwujiu:
## Descriptions
Need a new route that allows the upload of the ethics file.

Create a new endpoint `POST /file/ethics/{applicationId}` that receives a multi

Accepted ethics file types are PDF, DOC, DOCX. Maximum allowed file size is 5 MB. 

This endpoint will rely on a new function in the Application Controller that will save ethics letter. This controller relies on the creation of a [File Service](#208).

## Acceptance Criteria
- [x] Endpoint is created to allow upload of the ethics file
- [x] Express middleware restricts upload size to 5MB
- [x] Before saving the file:
  - [x] it is validated that the file is a pdf, doc, or docx (validate filename extension)
  - [x] ethics file changes can only be made when the application is in a state where they could edit their application
- [x]  Application controller function to save ethics letter should:
  - [x] In a single database transaction:
    - if this application has no ethics letter:
      - [x] create a new file entry and set the content to be the uploaded file
      - [x] set the application content ethics_letter property to be the ID of the new ethics file
    - else this application has a previous ethics letter:
      - [x] replace the contents of the file record (don't store multiple files for a single application).

### Auth Rules
For future reference once an auth system is in place:
- [ ] only the applicant can upload an ethics letter for their application.

JamesTLopez:
Note: 
This endpoint covers both upload new file and replace current file

Buwujiu:
Tested in dev, I was able to upload files on the Ethics page. Thanks James!

PR #277, opened 2025-03-06
#230, #207: File Upload Service With Ethics File Upload 

JamesTLopez:
## Summary

File upload service implementations and Ethics file upload functionality

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/230
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/207

## Description of Changes

- Create `/file/ethics/:applicationId` endpoint
- Create uploadEthicsFile Controller in applicationController file
- Create File Upload Service
- Add Test-Cases for applicationController and FileService
- Create middleware in `request-utils` to check for valid upload's

### Special Instructions
Before running these changes, you will need to install [formidable](https://www.npmjs.com/package/formidable) package and alterations to the `request-utils`:
```
pnpm i && pnpm build:all
```

### Postman or Bruno Endpoint
```
http://localhost:3000/applications/file/ethics/:id
and  make sure you go into the body, change it into MultiPartForm and the key for the form should be named: file. 

Should look something like this:

| key  | Value            |   |   |   |
|------|------------------|---|---|---|
| file | selectedfile.tsx |   |   |   |
|      |                  |   |   |   |
|      |                  |   |   |   |


```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
image

- [x] **Updated Tests**
- [x] Documentation


surajgoraya:
Given that the file upload service has its own service layer & adds quite a bit of functionality, I feel like maybe it should live in its own controller and router?

Perhaps something like `/files/` as a API endpoint? We did this for `/signature/` for a similar reason

JamesTLopez:
> Given that the file upload service has its own service layer & adds quite a bit of functionality, I feel like maybe it should live in its own controller and router?
> 
> Perhaps something like `/files/` as a API endpoint rather than relying on `/applications/`? We did this for `/signature/` for a similar reason

I was following the Acceptance criteria written up [here](https://github.com/Pan-Canadian-Genome-Library/daco/issues/207). It makes perfect sense with it being in its separate controller, but I can see why its in the application controller since its technically an Application related feature(and no where else). Unless we are expanding the scope of where the user can upload files, then I think its fine in the apps controller but if everyone agrees to changing to your suggestion i'm okay with that too.

@demariadaniel @Azher2Ali  What do you guys think?

surajgoraya:
I get where you're coming from, to be fair, I originally viewed the `/signature/` route the same way since it literally lives in the `application_contents` table & is only related to applications (the original API path was even `/applications/edit/signature`). But it was brought up that it'd be a good idea organization wise for it to be separated out since it had a fair amount of it's own logic.

I think we should separate stuff like files, signature, and collaborators because despite being intrinsically linked to applications, they all have a fair amount of logic which feels like it expands the scope of the applicationsController. 

I think this should especially be the case for files and collaborators since both these elements live in their own tables, so mixing this logic into the applicationsController & router feels confusing imo.

Curious to hear what @demariadaniel and @Azher2Ali think though!

apps/api/src/routes/applicationRouter.ts
```diff
@@ -284,57 +284,85 @@ applicationRouter.post('/reject', jsonParser, async (req, res) => {
...
-		async (request: Request, response: Response) => {
-			const { applicationId } = request.params;
-
-			if (!applicationId || !isPositiveInteger(Number(applicationId))) {
-				response.status(400).json({
-					message: 'Invalid request. ApplicationId is required and must be a valid number.',
-					errors: 'MissingOrInvalidParameters',
-				});
-			}
+	'/file/ethics/:applicationId',
```
demariadaniel, line 291: I think you're missing a closing `)` here

apps/api/src/routes/applicationRouter.ts
```diff
@@ -284,57 +284,85 @@ applicationRouter.post('/reject', jsonParser, async (req, res) => {
...
-						data: result.data,
+		if (result.success) {
+			res.status(200).send(result.data);
+			return;
+		} else {
+			const errorReturn = { message: result.message, errors: String(result.errors) };
+			res.status(500).send(errorReturn);
+			return;
+		}
+	}),
```
demariadaniel, line 313: I think you're missing a closing `)` here
(posted twice, first time was wrong line)

apps/api/src/routes/applicationRouter.ts
```diff
@@ -284,57 +284,85 @@ applicationRouter.post('/reject', jsonParser, async (req, res) => {
...
-				response.status(500).send({
-					message: 'Internal server error.',
-					errors: String(error),
-				});
-			}
-		},
+			},
+		),
 	),
 );
```
demariadaniel, line 367: And then something may be off with number of brackets here

JamesTLopez: Same as above

JamesTLopez:
> I get where you're coming from, to be fair, I originally viewed the `/signature/` route the same way since it literally lives in the `application_contents` table & is only related to applications (the original API path was even `/applications/edit/signature`). But it was brought up that it'd be a good idea organization wise for it to be separated out since it had a fair amount of it's own logic.
> 
> I think we should separate stuff like files, signature, and collaborators because despite being intrinsically linked to applications, they all have a fair amount of logic which feels like it expands the scope of the applicationsController.
> 
> I think this should especially be the case for files and collaborators since both these elements live in their own tables, so mixing this logic into the applicationsController & router feels confusing imo.
> 
> Curious to hear what @demariadaniel and @Azher2Ali think though!

Gotcha, ill make another controller for files!  🍡

demariadaniel:
> I get where you're coming from, to be fair, I originally viewed the `/signature/` route the same way since it literally lives in the `application_contents` table & is only related to applications (the original API path was even `/applications/edit/signature`). But it was brought up that it'd be a good idea organization wise for it to be separated out since it had a fair amount of it's own logic.
> 
> I think we should separate stuff like files, signature, and collaborators because despite being intrinsically linked to applications, they all have a fair amount of logic which feels like it expands the scope of the applicationsController.
> 
> I think this should especially be the case for files and collaborators since both these elements live in their own tables, so mixing this logic into the applicationsController & router feels confusing imo.
> 
> Curious to hear what @demariadaniel and @Azher2Ali think though!

Yes Files and Collaborators both have DB tables, need CRUD functionality and are represented in the data model so I think it's a fair organization approach to have separate services/controllers/routes for each.
Signatures is kind of a grey area in that it's updating a specific part of ApplicationContents. In this case I think it's separate enough that it represents it's own 'resource' for the time being.

We could potentially reorganize such that files/collaborators/signatures are sub-paths of Application.
Similar to Clinical: https://clinical.argo-qa.cumulus.genomeinformatics.org/api-docs/
See Sample Registration and Submission are prefixed by `/submission/program/{:programId}`
The routes / swagger updates wouldn't be too big a lift, would need some Zod reorg.

apps/api/src/controllers/fileController.ts
```diff
@@ -0,0 +1,96 @@
...
+		if (!canEditResult) {
+			return failure('Invalid action, must be in a draft state', 'Invalid action');
+		}
+
+		const ethicsLetterId = application.contents?.ethics_letter;
+
+		const txResult = await database.transaction(async (tx) => {
+			let result;
+
+			if (ethicsLetterId && ethicsLetterId !== null) {
```
demariadaniel, line 62: I would move this conditional out of `txResult`, seems like the kind of error you want to screen for before initiating anything

apps/api/src/service/fileService.ts
```diff
@@ -0,0 +1,125 @@
...
+const filesSvc = (db: PostgresDb) => ({
+	createFile: async ({
+		file,
+		application,
+		type,
+		transaction,
+	}: {
+		file: formidable.File;
+		application: JoinedApplicationRecord;
+		type: 'SIGNED_APPLICATION' | 'ETHICS_LETTER';
```
demariadaniel, line 43: This `type` param is worth capturing in its own variable (`/service/types` or `data-model`? where ever you see fit)

apps/api/src/controllers/fileController.ts
JamesTLopez: I need the response in both scenarios "create/update" file so I update the "ethics_letter" id on the application contents. If I pulled it out, i would have to rewrite the `applicationRepo.editApplication` functionality twice for both scenarios. Unless i'm missing something.

demariadaniel: I see what's happening now. So actually `createFile` needs a transaction argument.

Also L84 check is unneeded b/c you have `return result` at both L85 + L87

apps/api/tests/api/application-api.test.ts
```diff
@@ -175,29 +174,29 @@ describe('Application API', () => {
 		});
 	});
 
-	describe('Get Application Metadata', () => {
-		it('should get the counts for each of the application states', async () => {
-			const applicationRecordsResult = await testApplicationRepo.listApplications({ user_id });
+	// describe('Get Application Metadata', () => {
```
demariadaniel, line 177: We were discussing this test offline but I don't think it's been an issue in the pipeline? Don't need to commit this I think?

packages/request-utils/src/validation.ts
```diff
@@ -101,4 +102,61 @@ function withParamsSchemaValidation<ReqParams>(
...
+						.send({ message: 'Invalid file upload, file must be less than 5mb and only 1 file can be uploaded' });
+					return;
+				}
+
+				if (!files.file || !files.file[0]) {
+					response.status(400).send({ message: 'File does not exist' });
+					return;
+				}
+
+				const uploadedFile = files.file[0];
```
demariadaniel, line 138: Does this cover the case where it's `files.file` ? Referring to the conditional line 133

apps/api/src/controllers/fileController.ts
JamesTLopez: > I see what's happening now. So actually `createFile` needs a transaction argument.
> 
> Also L84 check is unneeded b/c you have `return result` at both L85 + L87

I think applicantResult should be returned there actually.

apps/api/tests/api/application-api.test.ts
JamesTLopez: Oh whoops, ill add it back in, i commented it in on acciedent

packages/request-utils/src/validation.ts
JamesTLopez: It should, that condition checks if files.file exist at all. If I remove !files.file, const uploadedFile would be possibly undefined.

demariadaniel: Sorry I worded that confusingly 

I meant to ask if `uploadedFile` is ever defined as `const uploadedFile = files.file`. But I see `file.file` is either `[] | undefined`

demariadaniel (approved):
Looking great 👍

Issue #232, opened 2025-02-20
Endpoint to Delete File By ID


Buwujiu:
## Descriptions
Delete File by ID - Remove file entry by ID. Note: should add a warning to the TSDoc for this that describes how "this will fail when references to this file haven't been removed (caller will need to remove the reference to this file from application content before deleting file).

edsu7:
Tested delete end point - works as expected

Notes:
Returns `204`, is `200` preferred?
![Image](https://github.com/user-attachments/assets/beee7ba8-1326-4792-90b8-3063ff1806f9)

JamesTLopez:
> Tested delete end point - works as expected
> 
> Notes: Returns `204`, is `200` preferred? ![Image](https://github.com/user-attachments/assets/beee7ba8-1326-4792-90b8-3063ff1806f9)

204 here because we do not return any content after deleting the file. If we were to return content in from the response, such as the id of the file deleted, then 200 would be the more correct solution. Either solution imo is correct, this is just technically more correct.

Buwujiu:
Thanks both @edsu7 @JamesTLopez ! Closing ticket!

PR #283, opened 2025-03-12
#232: Delete File By ID

JamesTLopez:
## Summary

Backend implementation of Delete File By ID

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/232

## Description of Changes

- Create Delete file endpoint
- Create `deleteFile` controller
- Create `deleteFileById` service
   - added transaction to handle delete CRUD

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/routes/fileRouter.ts
```diff
@@ -52,4 +52,23 @@ fileRouter.post(
 	}),
 );
 
+fileRouter.delete('/:fileId', async (req: Request, res: Response) => {
+	const { fileId } = req.params;
```
demariadaniel, line 56: Could add `withReqParams` here

apps/api/src/resources/swagger.yaml
```diff
@@ -651,6 +651,80 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ServerErrors'
+        '400':
+          description: 'Invalid fileId.'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ServerErrors'
```
demariadaniel, line 727: This should be a ClientError response

apps/api/src/resources/swagger.yaml
```diff
@@ -651,6 +651,80 @@ paths:
...
+        '400':
+          description: 'Error uploading file. File must be less than 5mb and only 1 file can be uploaded.'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ServerErrors'
```
demariadaniel, line 696: Also ClientError

apps/api/src/controllers/fileController.ts
```diff
@@ -95,3 +95,46 @@ export const uploadEthicsFile = async ({ applicationId, file }: { applicationId:
...
+
+			if (!deleteResult.success) {
+				return deleteResult;
+			}
+			const { application_id } = deleteResult.data;
+
+			const editApplicationResult = await applicationRepo.editApplication({
+				id: application_id,
+				update: {
+					ethics_letter: null,
```
demariadaniel, line 121: Could this be a different file type other than ethics_letter?

apps/api/src/resources/swagger.yaml
surajgoraya: If you're adding `withReqParams` you can use the following `oneOf` syntax to cover both potential ClientError objects and RequestValidationError objects:
```yaml
#....
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/responses/RequestValidationError'
                  - $ref: '#/components/responses/ClientErrors'
```

apps/api/src/controllers/fileController.ts
JamesTLopez: The file table can store two different types of files, "SIGNED_APPLICATION" or "ETHICS_LETTER".
Signed application has a different row in application contents which both point towards their respective id's in the files table

demariadaniel: I'm asking because I don't understand why `ethics_letter` is always `null` here; depending on the file ID you could have deleted another file type

JamesTLopez: AHH yes your right, there should be a check for the file type and switch which values needs to be reset to null.

apps/api/src/routes/fileRouter.ts
```diff
@@ -52,4 +53,26 @@ fileRouter.post(
...
+
+		if (!isPositiveInteger(id)) {
+			res.status(400).send({ message: 'Invalid fileId' });
+			return;
+		}
+
+		const result = await deleteFile({ fileId: id });
+
+		if (!result.success) {
+			res.status(500).send(result);
```
demariadaniel, line 70: There should be a `return` here after `send`

demariadaniel (approved):

Issue #260, opened 2025-03-04
Integrate Add Collaborator API with UI


Buwujiu:
## Acceptance Criteria

![Image](https://github.com/user-attachments/assets/20ba537c-cb8b-4ec0-8612-097792b1e34c)
- When applicants click on Add Collaborator, modal should close, pop up banner to indicate success or failure. 
- Added collaborator should appear in the collaborator table.

edsu7:
Tested the follow
|Test|Result|
|--|--|
|Adding collaborator works on dev page|Success|

Notes:
- Adding the same donor results in an unspecified error

![Image](https://github.com/user-attachments/assets/a1f8cad1-ab4e-4695-8a03-14efe0bab878)
while API response is :
```
{
    "message": "Cannot create duplicate collaborator records",
    "errors": "DuplicateRecords"
}
```

Buwujiu:
Thanks for testing! Logged one feedback in https://github.com/Pan-Canadian-Genome-Library/daco/issues/322. It's error msg update, doesn't affect the scope of this ticket, closing!

PR #284, opened 2025-03-12
#260: Add Collaborators API/UI integration

JamesTLopez:
## Summary

Integrate the Add Collaborators Endpoint with frontend using React-Query.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/260

### UI
- Remove Mock table row data
- Update table types to collaborator response type
- Create custom hook `useAddCollaborator` for react-query 
- Integrate add collaborator endpoint with react-query
- Create new folders for mutations/queries to separate different types of requests
- integrate hook into Add Collaborator modal
- update cache of Get Collaborators to include newly added collaborator if successful

### Server
- add `aliasCollaboratorRecord ` aliasing to create collaborator response object
- Update test-cases to be compatible with `aliasCollaboratorRecord `

### Special Instructions
Before running these changes run the following command in top level folder:
```
pnpm i && pnpm run build:all
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/ui/src/api/apiUtils.ts
```diff
@@ -0,0 +1,52 @@
...
+				error.errors = translate('errors.http.404.message');
+				break;
+			case 500:
+				error.message = translate('errors.http.500.title');
+				error.errors = translate('errors.http.500.message');
+				break;
+		}
+
+		throw error;
+	}
```
demariadaniel, line 51: This streamlines a ton of duplicated code! I think we can take implementation one step further so that it's closer to:
```
return await response.withErrorHandler(response, translate).json();
```

JamesTLopez: Im am unsure or unfamiliar with the pattern of adding `.withErrorHandler` custom method to the api response.
Are you talking about something like this?

```js
    const response = await fetch('/endpoint');
    const handledResponse = await withErrorHandler(response, translate);
    return await handledResponse.json();

```

demariadaniel: Discussed on Slack -- first suggestion was a half formed idea. My intent was more as a promise then/catch pattern so that `response` passes through handler and is returned when `response.ok`

`const response = await fetch('/endpoint').then(withErrorHandler);`

Easy to make this a standard addition on future `fetch`'s

demariadaniel (approved):
👏👏

Issue #178, opened 2025-01-28
Test Updates + API Refactors


demariadaniel:
Several updates which rose out of work related to: https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=85081519&issue=Pan-Canadian-Genome-Library%7Cdaco%7C3 
and will facilitate future development on the API

Potentially worth breaking into multiple tickets

1. Refactor sequential tests
Separate sequential state manager tests so that each test runs independently (does not depend on outcome of previous test to pass)
Sequential test can create extra work by linking multiple functions together

See: https://github.com/Pan-Canadian-Genome-Library/daco/pull/153

2. Mock Database Calls / Separate Unit Tests from Integration Tests
Known issue with Test Containers / Docker / Jenkins prevents running our current test suite in our pipelines for code change validation
Duplicate existing test files and convert to mocked unit tests where applicable

Relates to: https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=94825563&issue=Pan-Canadian-Genome-Library%7Cdaco%7C170

3. Refactor Service layer dependency injection
Create central dependency injection following example here: https://github.com/overture-stack/lyric/blob/main/packages/data-provider/src/core/provider.ts

function wrapper with dependencies, init services once, return object with functions
Replace current usage of `getDbInstance` and `service(db)`

surajgoraya:
Somewhat related, but as a stop-gap measure while we wait for automated integration testing, we should update the PR template to require a screenshot of the test suite passing

demariadaniel:
#2: Pipeline tests are working, but mocked requests/responses for testing server routes will be helpful
#4: Transaction pattern has been established (see stateManager._updateRecords / applicationRepo.findOneAndUpdate) and just needs to be extended to all services
#5: Much of this work has been done but there are several endpoints that should be updated to use `withSchemaValidation`

Buwujiu:
Some more dev improvement to be done, won't block anything, will get it done in the next sprint.

Buwujiu:
note: only 4. Review shared DB transactions in stateManager changes functionality, the rest do not.

Buwujiu:
This is not an urgent task right now, it needs breakdown to smaller tasks, we will park on it and get back to it later.

PR #285, opened 2025-03-12
#178 🧽 Chore: Refactor Alias functions

demariadaniel:
## Summary

Use ` ts-case-convert`with `zod` to alias snake_case/camelCase keys, with field parsing and validation
~Use Lodash `snakeCase` and `camelCase` to create typed functions for aliasing keys from one convention to another~

### Related Issues

- #178 

## Description of Changes
Creates reusable utility functions so aliasing object keys no longer has to be done manually
Rename `/utils/routes` to `/utils/aliases` b/c those are the only functions in this file
All refactored records have been tested using `lodash.isEqual`

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,142 @@
...
+		const value = data[key];
+		const accumulator = { ...acc, [aliasedKey]: value };
+		return accumulator;
+	}, {}) as DatabaseRecord;
+
+	return databaseData;
+};
+
+/** Convenience function for specific alias utils input/output scenarios */
+export const aliasApplicationRecord = (data: JoinedApplicationRecord): ApplicationResponseData => {
```
demariadaniel, line 94: This is an example where you're going from a complete database record to a partial UI record and so some of the keys are omitted

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,196 @@
...
+			applicationId: value.application_id,
+			collaboratorFirstName: value.first_name,
+			collaboratorMiddleName: value.middle_name,
+			collaboratorLastName: value.last_name,
+			collaboratorInstitutionalEmail: value.institutional_email,
+			collaboratorPositionTitle: value.position_title,
+			collaboratorPrimaryAffiliation: value.title,
+			collaboratorResearcherProfileURL: value.profile_url,
+			collaboratorSuffix: value.suffix,
+			collaboratorType: value.collaborator_type,
```
demariadaniel, line 191: The issue here is the keys are actually different, the DTO uses `collaboratorFirstName` vs the db which is simply `first_name`

apps/api/src/resources/swagger.yaml
```diff
@@ -980,7 +980,7 @@ components:
         update:
           description: Changes to make to the application
           type: object
-          $ref: '#/components/schemas/ApplicationContents'
+          $ref: '#/components/responses/ApplicationContentsResponse'
```
demariadaniel, line 983: This `ApplicationUpdates` is a request schema but the original `ApplicationContentsResponse` was in responses

Both incoming/outgoing data is using camelCase

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,196 @@
...
+const filterOmittedKeys = (key: string, omittedKeys: string[]) => !omittedKeys.includes(key);
+
+/**
+ * Helper function to convert Postgres snake_case to FE camelCase
+ * Generics allow usage w/ multiple input/output combinations
+ * @param data Original Database record with snake case keys to convert
+ * @param omittedKeys List of keys to remove from output for cases where a partial record is returned
+ * @returns ResponseRecord - Generic type representing new record with updated camelCase keys
+ */
+export const aliasToResponseData = <
```
demariadaniel, line 48: Open to changing the names, `responseData` / `databaseData` are maybe too narrow / not accurately descriptive enough

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,196 @@
...
+	DatabaseRecord extends Record<string, any>,
+	ResponseRecord extends Record<string, any>,
+>(
+	data: DatabaseRecord,
+	omittedKeys: string[] = [],
+): ResponseRecord => {
+	type snakeCaseKey = string & keyof DatabaseRecord;
```
ciaranschutte, line 55: can move the type outside of the function

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,196 @@
...
+	const allKeys = Object.keys(data).map((key): snakeCaseKey => key);
```
ciaranschutte, line 56: I don't think there's a need for the extra loop with `map`.. looks like maybe its a type fix of some kind but it's likely doable without the extra map...pretty much just casting the type anyway

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,196 @@
...
+ * @returns DatabaseRecord - Generic type representing new record with updated snake_case keys
+ */
+export const aliasToDatabaseData = <
+	RequestRecord extends Record<string, any>,
+	DatabaseRecord extends Record<string, any>,
+>(
+	data: RequestRecord,
+	omittedKeys: string[] = [],
+): DatabaseRecord => {
+	type camelCaseKey = string & keyof RequestRecord;
```
ciaranschutte, line 86: type in function

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,196 @@
...
+	data: DatabaseRecord,
+	omittedKeys: string[] = [],
+): ResponseRecord => {
+	type snakeCaseKey = string & keyof DatabaseRecord;
+	const allKeys = Object.keys(data).map((key): snakeCaseKey => key);
+
+	const filteredKeys = allKeys.filter((key) => {
+		return filterOmittedKeys(String(key), omittedKeys);
+	});
+
```
ciaranschutte, line 61: some overly typed `keys` I think, going back and forth because a strongly typed key and then converting it back into a regular string with `String`

I like `filterOmittedKeys` a lot. Could even make it a tad cleaner with something _like_
```javascript
const filterOmittedKeys = (key: string, omittedKeys: string[]) => (key) => !omittedKeys.includes(key);

const filteredKeys = allKeys.filter(filterOmittedKeys(String(key), omittedKeys));

**Review by ciaranschutte — COMMENTED** [review]
couple of comments:
- naming these more explicitly might make sense. `convertToApplicationRecord` rather than `alias`. Alias to me sounds like just renaming, but these objects are being changed, data transfer layer rather than just renaming.
- if you do `import { foo } from 'lodash'` it should tree shake the individual packages you need. using `import lodash from 'lodash'` imports the entire lodash lib and then calls individual functions

**Comment by demariadaniel** [comment]
> * if you do `import { foo } from 'lodash'` it should tree shake the individual packages you need. using `import lodash from 'lodash'` imports the entire lodash lib and then calls individual functions

Yes we discussed this in Slack as well but import was not working as expected, I will take another look

File: apps/api/src/utils/aliases.ts (hunk 6 above)
Comment (demariadaniel) [reply apps/api/src/utils/aliases.ts:56]: Yeah it's this or `as`, this is b/c Object.keys returns string[], I'll look up alternate solutions

**Comment by ciaranschutte** [comment]
> > * if you do `import { foo } from 'lodash'` it should tree shake the individual packages you need. using `import lodash from 'lodash'` imports the entire lodash lib and then calls individual functions
> 
> Yes we discussed this in Slack as well but import was not working as expected, I will take another look

Strange days. Not sure where the slack chat is but if you'd like a hand debugging let me know 👍

File: apps/api/src/utils/aliases.ts [hunk 9]
```diff
@@ -0,0 +1,196 @@
...
+		state,
+		createdAt,
+		approvedAt,
+		updatedAt,
+		expiresAt,
+		contents,
+	};
+};
+
+/** Convenience function for specific alias utils input/output scenarios */
```
demariadaniel, line 143: I'll restore comments that are in main, removed some of the jsdocs here

apps/api/src/utils/aliases.ts
ciaranschutte: closer to const filteredKeys = allKeys.filter(filterOmittedKeys(omittedKeys)) ;) I missed it first try

apps/api/src/utils/aliases.ts
demariadaniel: OK made multiple updates related to your feedback
- imports
- move type definition
- remove map and use of String
- refactor filterOmittedKeys

demariadaniel: I am going to leave this unresolved b/c the imports are indeed still an issue

demariadaniel: Something about this Lodash import bit doesn't work as advertised

Using `import lodash from 'lodash/snakeCase';` yields `Cannot find module 'lodash/snakeCase'`
Using `require` as suggested in the Docs yields an ESM error pushing you to use `import` instead

This is an API function so I don't think we need to be super concerned about dependency size here
So I've reverted to what works `import lodash;` `lodash.camelCase()` and sticking with that

demariadaniel: and have renamed the two core utility functions to `convertToCamelCase` and `convertToSnakeCase` so it's extremely clear what they do

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,195 @@
...
+>(
+	data: SnakeCaseRecord,
+	omittedKeys: string[] = [],
+): CamelCaseRecord => {
+	const allKeys: inputKey[] = Object.keys(data);
+
+	const filteredKeys = allKeys.filter(filterOmittedKeys(omittedKeys));
+
+	const responseData = filteredKeys.reduce((acc, key) => {
+		const aliasedKey = lodash.camelCase(key);
```
JamesTLopez, line 62: is it possible to tree shake this?

demariadaniel: See this discussion and elsewhere in the PR, this isn't working as expected for me.
https://github.com/Pan-Canadian-Genome-Library/daco/pull/285#discussion_r1994217410

I can get this to a state where the import works fine in `utils/aliases`, the app builds, then the test runner fails. So rather than wrangle with it more I just chose to use the whole import.

apps/api/tests/functions/alias-utils.test.ts
```diff
@@ -0,0 +1,79 @@
...
+			created_at: new Date(),
+			approved_at: new Date(),
+		};
+
+		const aliasResult = convertToCamelCase<ApplicationModel, ApplicationDTO>(testData);
+
+		assert.ok(aliasResult.hasOwnProperty('userId'));
+		assert.ok(aliasResult.hasOwnProperty('createdAt'));
+		assert.ok(aliasResult.hasOwnProperty('approvedAt'));
+	});
```
joneubank, line 59: These tests appear to be duplicate.

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,197 @@
...
+	const allKeys: inputKey[] = Object.keys(data);
+
+	const filteredKeys = allKeys.filter(filterOmittedKeys(omittedKeys));
+
+	const databaseData = filteredKeys.reduce((acc, key) => {
+		const aliasedKey = snakeCase(key);
+		const value = data[key];
+		const accumulator = { ...acc, [aliasedKey]: value };
+		return accumulator;
+	}, {}) as SnakeCaseRecord;
```
joneubank, line 96: Don't cast types! This will tell typescript that the returned object is whatever type you declare in the generic, even if that is incorrect.

```ts
type Camel = { camelCase: string };
type Snake = { snake_case: string };
const output = convertToSnakeCase<Camel, Snake>({ camelCase: 'asdf' });
output.snake_case; // doesn't exist but TS thinks it does
```

apps/api/src/utils/aliases.ts
```diff
@@ -0,0 +1,197 @@
...
+
+	return responseData;
+};
+
+/**
+ * Helper function to convert FE camelCase to Postgres snake_case
+ * Generics allow usage w/ multiple input/output combinations
+ * @param data Original Request record with camelCase keys to convert
+ * @param omittedKeys List of keys to remove from output for cases where a partial record is returned
+ * @returns DatabaseRecord - Generic type representing new record with updated snake_case keys
```
joneubank, line 78: This doesn't return a generic type. This returns the object you describe here.

joneubank (changes requested):
The case conversion you have here will work provided no one makes a mistake with the provided generic types, but this isn't true type safety.

I found an existing library that doesnt this while preserving type information: https://www.npmjs.com/package/ts-case-convert


Digging into the source code, its actually quite tricky to perform the key conversion in the type system (tricky, but possible and not too hard to read):
https://github.com/RossWilliams/ts-case-convert/blob/main/src/caseConvert.ts
Take a look at the type `ToCamel` and `ObjectToCamel` for examples.

In addition to the type safety, this also recursively updates all properties of the object, not just the top level keys. Would you consider using this library or similar instead of our custom methods?

apps/api/package.json
```diff
@@ -15,10 +15,11 @@
 		"migrate": "pnpm drizzle-kit migrate",
 		"start:prod": "pnpm migrate && tsx ./src/main.js",
 		"test": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/**/*.test.ts",
-		"test:local": "pnpm run test:api && pnpm run test:db && pnpm run test:services",
+		"test:local": "pnpm run test:api && pnpm run test:db && pnpm run test:services && pnpm run test:functions",
 		"test:api": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/api/*.test.ts",
 		"test:db": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/db/*.test.ts",
-		"test:services": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/services/*.test.ts"
+		"test:services": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/services/*.test.ts",
+		"test:functions": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/functions/*.test.ts"
```
demariadaniel, line 22: In the test updates PR I would like to use a `test:regex` approach rather than continually adding individual scripts

apps/api/src/utils/aliases.ts
demariadaniel: This comment (and function) have been removed

apps/api/src/utils/aliases.ts
demariadaniel: This has been removed, updated to use `ts-case-convert` with `zod` schemas

apps/api/tests/functions/alias-utils.test.ts
demariadaniel: Rewrote all unit tests

demariadaniel:
> The case conversion you have here will work provided no one makes a mistake with the provided generic types, but this isn't true type safety.
> 
> I found an existing library that does this while preserving type information: https://www.npmjs.com/package/ts-case-convert
> 
> Digging into the source code, its actually quite tricky to perform the key conversion in the type system (tricky, but possible and not too hard to read): https://github.com/RossWilliams/ts-case-convert/blob/main/src/caseConvert.ts Take a look at the type `ToCamel` and `ObjectToCamel` for examples.
> 
> In addition to the type safety, this also recursively updates all properties of the object, not just the top level keys. Would you consider using this library or similar instead of our custom methods?

Major update to use `ts-case-convert` in combination with `zod` schemas / `@pcgl-daco/validation`

apps/api/package.json
```diff
@@ -41,15 +43,18 @@
...
+		"connect-redis": "^8.0.1",
 		"cors": "^2.8.5",
 		"drizzle-orm": "^0.39.1",
-		"connect-redis": "^8.0.1",
+		"drizzle-zod": "^0.7.0",
 		"express": "^4.21.0",
 		"express-session": "^1.18.1",
 		"formidable": "^3.5.2",
 		"ioredis": "^5.5.0",
+		"lodash": "^4.17.21",
```
JamesTLopez, line 57: Do we need lodash still? If not we can remove this and the associated types

apps/api/tests/functions/alias-utils.test.ts
```diff
@@ -0,0 +1,87 @@
...
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { JoinedApplicationRecord } from '@/service/types.ts';
```
JamesTLopez, line 20: Is this `type JoinedApplicationRecord` ?

apps/api/package.json
demariadaniel: That should be removed now, let me see what happened

demariadaniel: Yeah see the comment is now marked as `Outdated`, maybe refresh and see if the diff is updated on your side? It is for me

Removed this yesterday: https://github.com/Pan-Canadian-Genome-Library/daco/commit/69c54fd6fc5df132c0a1ef94a256122f62f74a9b

I also encountered some merge w/ main weirdness you can see in the commits which may be related

JamesTLopez: Weird, I just opened this PR this morning, but it looks like its gone now. 🤔 
Oh well, its fixed now XD

apps/api/tests/functions/alias-utils.test.ts
demariadaniel: This is an `interface` but `UpdateEditApplicationRequest` needed the `type` import added

apps/api/src/utils/aliases.ts
joneubank: I'm fine with this name. Wondering if "shape" is correct vs "data"

joneubank:
> * naming these more explicitly might make sense. `convertToApplicationRecord` rather than `alias`. Alias to me sounds like just renaming, but these objects are being changed, data transfer layer rather than just renaming.

I also prefer `convert` to `alias`.

> * if you do `import { foo } from 'lodash'` it should tree shake the individual packages you need. using `import lodash from 'lodash'` imports the entire lodash lib and then calls individual functions

If this is only being used serverside on the API, I am not worried about tree shaking. This is a good lesson in general though.

joneubank (approved):

demariadaniel:
> > * naming these more explicitly might make sense. `convertToApplicationRecord` rather than `alias`. Alias to me sounds like just renaming, but these objects are being changed, data transfer layer rather than just renaming.
> 
> I also prefer `convert` to `alias`.

Updated to use `convertTo` naming convention

surajgoraya (approved):
changes lgtm, will be nice to finally have a good aliasing solution! :shipit:

Issue #123, opened 2024-12-18
Endpoint - Revoke Application


Buwujiu:
## Descriptions
Build a REST endpoint to allow applicant and DAC revoke application.
https://docs.google.com/drawings/d/1CI-nB8MBr6Abc6DlGCr2moKNFcfHaG_YjRU1uVk61Lw/edit
## Acceptance Criteria
- Can only take place when an application is Approved. 
- Revoked application cannot be re-opened.
- the app status will change from Approved to Revoked.
- Add unit test
- Add error handling

Buwujiu:
PR will be ready soon, by End of week.

Buwujiu:
Will test in dev using UI. Closing ticket!

PR #287, opened 2025-03-13
#123: Add REST endpoint for applicant and DAC to revoke application

Azher2Ali:
#123: Add REST endpoint for applicant and DAC to revoke application

## Summary

This PR introduces a new REST endpoint that allows an applicant or DAC member to revoke an application. Once revoked, an application cannot be reopened, and its status will change from Approved to Revoked.

## Description of Changes
Added `POST /applications/{applicationId}/revoke `endpoint.

- Only accessible if the application is in the Approved state.
- Changes the application's status from Approved to Revoked.
- Prevents revoked applications from being reopened.
- Includes proper error handling for invalid states or permissions.
- Updated OpenAPI documentation to include the new endpoint.
- Added unit and integration tests to validate the revocation logic.

**Error Handling**

- Returns 400 Bad Request if the application is not in Approved state.
- Returns 403 Forbidden if the requester lacks permission to revoke.
- Returns 404 Not Found if the application does not exist.
- Returns 500 Internal Server Error for unexpected failures.

### Special Instructions
```
pnpm build:all
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
  - The PR title is properly formatted to match the pattern: `#123: Add REST endpoint for applicant and DAC to revoke application`
- [x] **Labels Added**
  - Label is added for each package/app that is modified (`api`)
- [x] **Local Testing**
- [x] **Updated Tests**


apps/api/src/routes/applicationRouter.ts
```diff
@@ -350,6 +350,55 @@ applicationRouter.post(
 	),
 );
 
+applicationRouter.post('/applications/:applicationId/revoke', jsonParser, async (request: Request, response: Response) => {
```
JamesTLopez, line 353: You can use the `withParamsSchemaValidation` here.

apps/api/src/routes/applicationRouter.ts
```diff
@@ -350,6 +350,55 @@ applicationRouter.post(
...
+	const { applicationId } = request.params;
+
+	if (!applicationId || isNaN(parseInt(applicationId))) {
+		response.status(400).json({
+			message: 'Invalid request. ApplicationId is required and must be a valid number.',
+			errors: 'MissingOrInvalidParameters',
+		});
```
JamesTLopez, line 360: Should have a return here.

apps/api/src/routes/applicationRouter.ts
```diff
@@ -350,6 +350,55 @@ applicationRouter.post(
...
+			} else if (errors === 'StateConflict') {
+				status = 409;
+				message = 'Application is already revoked.';
+			} else if (errors === 'Unauthorized') {
+				status = 403;
+				message = 'Unauthorized to revoke this application.';
+			} else if (errors === 'InvalidState') {
+				status = 400;
+				message = 'Cannot revoke application in its current state.';
+			}
```
JamesTLopez, line 390: Up to you for this one, I think it will be easier to read if this was a switch case.

apps/api/src/routes/applicationRouter.ts
Azher2Ali: `withParamsSchemaValidation`requires schema for request body which this endpoint does not have. I did check other endpoints too, which are only using params and not the request body and those endpoints are not using `withParamsSchemaValidation`

apps/api/src/routes/applicationRouter.ts
Azher2Ali: Nice suggestion!

apps/api/src/routes/applicationRouter.ts
JamesTLopez: It is a more recent adopted pattern that we have been using. `withParamsSchemaValidation` is used on the `collaboratorsRouter` for a bunch of endpoints. You are correct, there is no schema so you would have to make one.

 But upon review, theres a bunch of endpoints that basically use the same param `:applicationId`  and the zod schema is identical. Perhaps we can have a common schema?
 
 @demariadaniel @Azher2Ali  @surajgoraya What do you guys think?

apps/api/src/routes/applicationRouter.ts
```diff
@@ -350,6 +350,61 @@ applicationRouter.post(
...
+	if (!applicationId || isNaN(parseInt(applicationId))) {
+		response.status(400).json({
+			message: 'Invalid request. ApplicationId is required and must be a valid number.',
+			errors: 'MissingOrInvalidParameters',
+		});
+		return;
+	}
+
+	try {
+		const applicationIdNum = Number(applicationId);
```
joneubank, line 365: I think this will be updated with the use of `withParamSchemaValidation`, but without that, would you consider defining `applicationIdNum` at the start of this block using `parseNum(applicationId || '')`. Its awkward converting it to test and then converting it again for use:

```ts
const applicationIdNum = parseInt(applicationId || '');
	if (isNaN(applicationIdNum)) {
		response.status(400).json({
			message: 'Invalid request. ApplicationId is required and must be a valid number.',
			errors: 'MissingOrInvalidParameters',
		});
		return;
	}
```

apps/api/src/routes/applicationRouter.ts
```diff
@@ -350,6 +350,61 @@ applicationRouter.post(
 	),
 );
 
+applicationRouter.post('/applications/:applicationId/revoke', jsonParser, async (request: Request, response: Response) => {
```
joneubank, line 353: I believe we have moved all the `/applications` paths out of this router, can we make this consistent with the other endpoints? (also needs to be removed from /close).

apps/api/src/controllers/applicationController.ts
```diff
@@ -257,6 +257,44 @@ export const submitRevision = async ({ applicationId }: { applicationId: number
...
+			return result;
+		}
+
+		const application = result.data;
+
+		const appStateManager = new ApplicationStateManager(application);
+
+		if (appStateManager.state !== ApplicationStates.APPROVED) {
+			return failure('Application should be in APPROVED status', 'RevokeConflict');
+		}
```
joneubank, line 277: All the logic for what states are valid to revoke from are stored in the state manage, we don't need this check, we only need the next line's `appStateManager.revokeApproval()`.

joneubank (approved):

PR #289, opened 2025-03-13
#178 🗑️ Chore Initial Test Updates

demariadaniel:
## Summary

Test suites organization improvements for efficiency & DX

### Related Issues

- [178](https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=95531183&issue=Pan-Canadian-Genome-Library%7Cdaco%7C178)

## Description of Changes
- Add individual `test:{folder}` scripts and change API `test` script to trigger each folder script instead of using glob
- Reorganize State Transition tests so that linked tests are grouped together & other tests are decoupled
- Revise List Applications DB population script approach
- Add folder for testUtils
- Remove `after(delete)` functions

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/package.json
```diff
@@ -14,7 +14,10 @@
 		"setup:env": "cp -n .env.schema .env",
 		"migrate": "pnpm drizzle-kit migrate",
 		"start:prod": "pnpm migrate && tsx ./src/main.js",
-		"test": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/**/*.test.ts"
+		"test": "pnpm run test:api && pnpm run test:db && pnpm run test:services",
+		"test:api": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/api/*.test.ts",
+		"test:db": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/db/*.test.ts",
+		"test:services": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/services/*.test.ts"
```
demariadaniel, line 20: This approach to `test:all` appears to have a made a performance improvement on my end, may resolve some of the issues w/ lengthy test runs that other team mates have encountered

apps/api/src/controllers/stateManager.ts
```diff
@@ -241,11 +241,21 @@ export class ApplicationStateManager extends StateMachine<ApplicationStateValues
...
 	}
 
 	async closeDacReview() {
-		return this.closeDraft();
+		const transitionResult = this._canPerformAction(close);
+		if (transitionResult.success) {
+			return await this._dispatchAndUpdateAction(close, 'close');
+		} else {
+			return failure(`Cannot close application with state ${this.getState()}`);
+		}
```
demariadaniel, line 258: The `close` functions all have the same functionality at the moment, but if they all call `closeDraft` then there's no separation btwn one `close` method and another

This change enables us to update the individual `close` functions as needed but they still have the core `close` action

apps/api/tests/api/application-state.test.ts
```diff
@@ -68,8 +65,12 @@ describe('State Machine', () => {
 		testApplicationRepo = applicationSvc(db);
 	});
 
-	describe('Application State Manager', () => {
+	describe('Success path: Create/Edit/Revisions/Submit/Approve', () => {
```
demariadaniel, line 68: Breaking state manger tests into 2 major groups

`success` path tests which all use one applicationRecord
`failure` path tests which use multiple records

apps/api/tests/api/application-state.test.ts
```diff
@@ -423,8 +445,6 @@ describe('State Machine', () => {
 	});
 
 	after(async () => {
-		await db.delete(applicationActions);
-		await db.delete(applications);
```
demariadaniel, line 427: Every test is using an individual test container so these `deletes` are likely just adding time and possibly a source of errors team members are encountering

apps/api/tests/utils/testUtils.ts
```diff
@@ -52,7 +52,7 @@ export const initTestMigration = async (db: PostgresDb) => {
 /** Sets Number of Db Records to Seed @ start of test run using addInitialApplications */
 export const numTestApplications = 20;
 /** Used to bypass pagination impacting test results */
-export const allRecordsPageSize = 2 * numTestApplications;
+export const allRecordsPageSize = 5 * numTestApplications;
```
demariadaniel, line 55: This number is just arbitrarily large so that we always get all records returned for pagination tests

apps/api/package.json
JamesTLopez: Question, Jenkins was running the test suite fine, should we just have a test:dev instead and keep the previous script?

JamesTLopez (approved):
Looks straight forward to me

apps/api/package.json
demariadaniel: The original intent was to debug the hanging test suite issue you guys are experiencing. So changing up the script seemed worth investigating.
There could be benefit to using a regex pattern rather than maintaining and updating a series of `pnpm test:a && pnpm test:b`: https://pnpm.io/cli/run#running-multiple-scripts

I misspoke when I said the tests were working in Jenkins -- there's still an issue with TestContainers and Jenkins. And the Jenkins Test stage isn't implemented.
We're going to resolve this with a separation for Unit Tests that removes TestContainers so it is able to run in the pipeline. Integration tests will still use TestContainers.
So it makes sense to start breaking up separate `test` scripts (i.e. `test:unit` vs `test:integration`) and change this `test:all` similarly.

Could go a couple ways, what do you think?
I think we could always revert to the previous script, but there's improvements to be made, unless you see any down sides to this set up.

demariadaniel: To clarify, the Tests do run in the pipeline build, for example: https://jenkins.sweng-dev.cumulus.genomeinformatics.org/blue/organizations/jenkins/softeng%2Fpcgl%2Fapp-builds%2Fdaco/detail/178%2Fchore-refactor-alias-functions/6/pipeline

It's just not a dedicated Test stage

demariadaniel: Yeah not sure about this, kind of premature optimization. 
The `&&` syntax does seem to produce better performance results based off the `duration_ms` results I see in the Node coverage report.

Thinking ahead, looking at the next branch, `test:unit:*` vs `test:integration:*` becomes useful. You don't want 16 different node scripts, i.e. `run test:integration:services` vs `run test:unit:services`. 
But ideally you could still trigger it that way as needed. I'm reviewing how you'd make that work.

JamesTLopez: Im okay with the new script (i like it alot actually), I was just wondering if it was worth keeping both so something like this:

`"test:local": "pnpm run test:api && pnpm run test:db && pnpm run test:services" ` //we'd run locally
`"test": "tsx --test --test-reporter=spec --experimental-test-coverage ./tests/**/*.test.ts"`

If the issue we were experiencing wasn't a problem on jenkins, then we could retain the script.

demariadaniel: Sure that's a good suggestion. Then we can compare/contrast the scripts without impacts on Jenkins yet. Follow up with larger updates in next PR.

demariadaniel: Added here: [Separate test / test:local script](https://github.com/Pan-Canadian-Genome-Library/daco/pull/289/commits/88f61bd697d5a56aa9959942cb9ec41a039530f3)

The `**/*` approach can be pretty performant but it also fluctuates! Sometimes slower, sometimes faster than the `test:local`. The `&&` script seems to be more consistent. Anyways. 🤓

JamesTLopez (dismissed):

JamesTLopez (approved):
Looks good to me! I like the new scripts. Runs fine on my end 👍

PR #325, opened 2025-03-31
#215: Application PDF Generation & API To Download Application PDF

surajgoraya:
### Related Issues

- Paste URL to issue 1
- Paste URL to issue 2

## Readiness Checklist

- [ ] **Self Review**
  - I have performed a self review of code
  - I have run the application locally and manually tested the feature
- [ ] **PR Format**
  - The PR title is properly formatted to match the pattern: `#{TicketNumber}: Description of Changes`
  - Links are included to all relevant tickets
- [ ] **Labels Added**
  - Label is added for each package/app that is modified (`api`, `ui`, `data-model`, etc.)
  - Label is added for the type of work done in this PR (`feature`, `fix`, `chore`, `documentation`)
- [ ] **Local Testing**
  - Successfully built all packages locally
  - Successfully ran all test suites, all unit and integration tests pass
- [ ] **Updated Tests**
  - Unit and integration tests have been added that describe the bug that was fixed or the features that were added
- [ ] Documentation
  - All new environment variables added to `.env.schema` file and documented in the README
  - All changes to server HTTP endpoints have open-api documentation
  - All new functions exported from their module have TSDoc comment documentation

PR #349, opened 2025-04-10
feat: 🔏 Add Auth Protection to `/files` & standardize Associated Rep Checks

surajgoraya:
## Summary

Adds auth protections to various `/files` endpoints, along with a standardized way to check if an application is associated to a particular Institutional Rep. It also refactors our unhandled exception stuff in `request-utils`. 

This PR was originally a part of #325 but that PR has gotten too big.

### Related Issues
- N/A

### 💻 Server
- 🌸  Added `isAssociatedRep` - a function which will check if a application is associated to a rep using auth. This for now always returns `true` but will be changed in the future.

#### 📁 Files API - Introduced proper auth protections. 🔏 
Files API did not have any auth protections, these have been applied:
- Added auth to `/download`, delete, upload and get.
- Changes `fileService.ts` to return appropriate errors depending on if a file was not found, rather than just returning SYSTEM_ERROR.

### 📦 Data Model
- 🔧  Modification to `FilesDTO `:
  - `submitterUserId` should be a string
  - `filename` should be optional 
  - Add `id`

### 💌 Request Utils:
- Changed `ErrorType` to return the style of errors we return in the API. Started using it in the APIs.
- Consolidated duplicated types for global error handling, removed uses of enums.

### 🌐 Global:
- 🌸 Prettier ran on the changes in `main`, sorry for the random files that look modified.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/routes/fileRouter.ts
```diff
@@ -17,110 +17,242 @@
...
+		!isApplicationInstitutionalRep ||
+		(userRole === 'APPLICANT' && userInfo?.userId !== result.data.submitterUserId)
+	) {
+		res.status(403).json({
+			error: ErrorType.FORBIDDEN,
+			message: 'Looks like you do not own, or have the rights to access to this file.',
+		});
 		return;
+	}
+
```
JamesTLopez, line 90: I think the logic here is a little off.
If the user is NOT an `isApplicationInstitutionalRep` but IS an applicant belonging to the file, it will still throw the auth error.  Also I don't think you need to check if the user role is `APPLICANT`, just checking if the ID's are the same should be enough, unless im missing something.

Something like this should work:

```ts
	const isApplicationInstitutionalRep = await isAssociatedRep(req.session, result.data.applicationId);

	if (isApplicationInstitutionalRep) {
		res.status(200).send(result.data);
                return
	}

	if (userInfo?.userId !== result.data.submitterUserId) {
		res.status(403).json({
			error: ErrorType.FORBIDDEN,
			message: 'Looks like you do not own, or have the rights to access to this file.',
		});
		return;
	}


```

apps/api/src/routes/fileRouter.ts
```diff
@@ -17,110 +17,242 @@
...
+import { getUserRole, isAssociatedRep } from '@/service/authService.ts';
 import { apiZodErrorMapping } from '@/utils/validation.ts';
-import { fileUploadValidation, withParamsSchemaValidation } from '@pcgl-daco/request-utils';
+import { FilesDTO } from '@pcgl-daco/data-model';
+import { ErrorType, fileUploadValidation, withParamsSchemaValidation } from '@pcgl-daco/request-utils';
 import { fileDeleteParamsSchema, getFileByIdParamsSchema, isPositiveInteger } from '@pcgl-daco/validation';
-import express, { type Request, type Response } from 'express';
+import express, { type Request } from 'express';
 import formidable from 'formidable';
+import { ResponseWithData } from './types.ts';
```
JamesTLopez, line 30: `type` here

JamesTLopez (commented):
Pretty solid, just that one logic change. 🔥

apps/api/src/controllers/fileController.ts
```diff
@@ -101,7 +102,17 @@ export const uploadEthicsFile = async ({
...
 			return result;
 		});
-		return txResult;
+		if (!txResult.success) {
+			return txResult;
+		}
+
+		const fileRecord = convertToFileRecord(txResult.data);
+
+		if (!fileRecord.success) {
```
demariadaniel, line 111: This conditional is redundant, both L112 & L115 return fileRecord

packages/request-utils/src/errors.ts
```diff
@@ -19,7 +19,10 @@
 
 export class BadRequest extends Error {
 	constructor(msg: string, details?: unknown) {
-		super(msg || "Sorry, looks like you sent a sent bad request. Please check the request and try again, or refer to our API documentation.");
+		super(
+			msg ||
+				'Sorry, looks like you sent a sent bad request. Please check the request and try again, or refer to our API documentation.',
```
demariadaniel, line 24: Typo `sent a sent`

packages/request-utils/src/responses.ts
```diff
@@ -18,19 +18,30 @@
...
 export const RequestValidationErrorResponse = <T>(
 	error: ZodError<T>,
 	customMessage?: string,
 ): RequestValidationError => ({
-	error: ErrorName.REQUEST_VALIDATION_ERROR,
-	message: customMessage ?? 'The request is invalid.',
+	error: ErrorType.INVALID_REQUEST,
+	message:
+		customMessage ??
+		'Sorry, looks like you sent a sent bad request. Please check the request and try again, or refer to our API documentation.',
```
demariadaniel, line 36: Typo here as well

apps/api/src/controllers/fileController.ts
surajgoraya: Good catch!

apps/api/src/routes/fileRouter.ts
```diff
@@ -17,110 +17,240 @@
...
+	return;
+}
+
+fileRouter.get(
+	'/:fileId',
+	authMiddleware({
+		requiredRoles: ['APPLICANT', 'DAC_MEMBER', 'INSTITUTIONAL_REP'],
 	}),
+	withParamsSchemaValidation(
+		getFileByIdParamsSchema,
```
leoraba, line 98: just noticed this schema error message mentions `applicationId`  instead of `fileId`.

apps/api/src/routes/signatureRouter.ts
```diff
@@ -174,7 +173,7 @@ signatureRouter.post(
 				const isApplicationUser = signee === 'APPLICANT' && applicationResult.data.userId === userId;
 				// TODO: Identify if the user role is institutional rep and is the rep for this application
```
leoraba, line 174: is there something else to do? otherwise remove the `TODO` from comment

leoraba (commented):
pretty good just small comments

apps/api/src/routes/signatureRouter.ts
surajgoraya: technically the `isAssociatedRep` function is always returning true because it's pending logic to actually be implemented properly. I can remove the `TOOD:` here since there's one in the `isAssociatedRep` file

apps/api/src/routes/fileRouter.ts
surajgoraya: Good catch!

apps/api/src/routes/fileRouter.ts
```diff
@@ -17,110 +17,240 @@
...
+	),
+);
+
+fileRouter.get(
+	'/:fileId/download',
+	authMiddleware({
+		requiredRoles: ['APPLICANT', 'DAC_MEMBER', 'INSTITUTIONAL_REP'],
+	}),
+	withParamsSchemaValidation(
+		fileDeleteParamsSchema,
```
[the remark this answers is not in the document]
surajgoraya: fixed! didn't notice this!

packages/request-utils/src/types.ts
```diff
@@ -19,20 +19,20 @@
...
+	FORBIDDEN: 'FORBIDDEN',
+	NOT_IMPLEMENTED: 'NOT_IMPLEMENTED',
 } as const;
 
-export type ErrorNames = (typeof ErrorName)[keyof typeof ErrorName];
+export type ErrorTypes = (typeof ErrorType)[keyof typeof ErrorType];
 
 export type ErrorResponse = {
-	error: ErrorNames;
+	error: ErrorTypes | 'NOT_IMPLEMENTED';
```
demariadaniel, line 35: Why `I 'NOT_IMPLEMENTED'` ? It's included in ErrorTypes

apps/api/src/routes/fileRouter.ts
```diff
@@ -17,110 +17,240 @@
...
+		},
+	),
 );
 
 fileRouter.post(
 	'/ethics/:applicationId',
-	fileUploadValidation(async (req: Request<any, { file: formidable.File }>, res: Response) => {
-		const { applicationId } = req.params;
+	authMiddleware({
+		requiredRoles: ['APPLICANT', 'DAC_MEMBER'],
```
demariadaniel, line 129: Double checking, can institutional rep access this endpoint?

demariadaniel: Same question for /delete

packages/request-utils/src/types.ts
surajgoraya: I added this while merging in previous changes that the other duplicate types file had, looking at it now, you're right, this isn't needed!

apps/api/src/routes/fileRouter.ts
surajgoraya: I don't believe Institutional Reps would be able to upload an Ethics file, given that this part of the application is specifically supposed to be completed by the Applicant, and the Rep shouldn't need any ability to upload and delete (though they can see the ethics file). 

Thus, I think it makes sense to give the ability to upload and delete to the DAC Members (who are treated as de facto admins), but not to Reps.

Perhaps by assumption here is wrong?

surajgoraya: @demariadaniel, just confirmed with @Buwujiu, Reps should not have permissions to do any ethics management, so this permissions structure should be good to go!

demariadaniel: Thanks I couldn't find a written reference to confirm and wanted to make sure 👍

demariadaniel (dismissed):

apps/api/src/routes/fileRouter.ts
surajgoraya: No worries! Thanks for the double check, the permissions structure being added after means we have to be careful! 🕺🏼

demariadaniel (approved):

Issue #273, opened 2025-03-05
Application Editing - Sign and Submit


JamesTLopez:
## Descriptions

Integrate Sign and Submit View with backend implementation.

## Acceptance Criteria

- [x] Side Menu navigation should be disabled until all sections are filled in
- [x] If a user signs the pdf, returns back to another section and triggers an invalid field and attempts to return back to the `Sign and Submit` section. The user is to be re-routed back to the `Introduction` section

Buwujiu:
Tested in dev, working as expected!

PR #354, opened 2025-04-15
#273: Sign and Submit Integration and Sidebar Changes

JamesTLopez:
## Summary

Integrate Sign and Submit page with API and update sidebar to check if all sections are filled before enabling sign and submit.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/273

## Description of Changes

- Create new react-query hooks:
   - useGetSignature
   - useCreateSignature
- Integrate sign and submit section with API and react-query hooks
- Create ValidateAllSections validator function to verify if ALL sections are completed
- Limit max size of signature box
- Re-order icon logic for sidemenu

## Readiness Checklist

- [ ] **Self Review**
- [ ] **PR Format**
- [ ] **Labels Added**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation

### DIFF
--- a/apps/ui/src/api/mutations/useCreateSignature.ts
+++ b/apps/ui/src/api/mutations/useCreateSignature.ts
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2025 The Ontario Institute for Cancer Research. All rights reserved
+ *
+ * This program and the accompanying materials are made available under the terms of
+ * the GNU Affero General Public License v3.0. You should have received a copy of the
+ * GNU Affero General Public License along with this program.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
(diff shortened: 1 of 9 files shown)


apps/ui/src/api/mutations/useCreateSignature.ts
```diff
@@ -0,0 +1,58 @@
...
+			return await response.json();
+		},
+		onError: (error) => {
+			notification.error({
+				message: error.message,
+			});
+		},
+		onSuccess: async (data) => {
+			// Invalidate previous signature get request
+			await queryClient.invalidateQueries({ queryKey: [`signature-${data.id}`], exact: false });
```
leoraba, line 53: is there a reason of using `exact: false` when passing a `queryKey`?

JamesTLopez: Ah this should be exact, nice catch! Making changes

apps/ui/src/api/mutations/useCreateSignature.ts
```diff
@@ -0,0 +1,58 @@
...
+					signature,
+					signee,
+				}),
+			}).then(withErrorResponseHandler);
+
+			return await response.json();
+		},
+		onError: (error) => {
+			notification.error({
+				message: error.message,
```
surajgoraya, line 48: This should likely use a standardized error message via the translations file, all of our error messages on the backend are English with no French conversion ability just yet

apps/ui/src/pages/applications/sections/sign.tsx
```diff
@@ -20,37 +20,59 @@
...
 		setOpenModal(true);
 	};
 
-	const onSaveClicked = () => {
-		console.log('saved');
+	const onSaveClicked = async () => {
+		const signature = getValues('signature');
+
+		if (signature && role) {
+			await createSignature({ applicationId: appId, signature, signee: 'APPLICANT' }).then(() => {
```
surajgoraya, line 70: Hmmm, two things, so ideally the API would determine the signee based on the userRole, we could include this change in this PR, or make it a separate one (separate might be better).

Second, this is hardcoding Applicant as the signee, we do need to have a signature field for the Rep. I think in our meeting we decided that it'd be the same signature field, and we wouldn't need to show the Applicant Signature? Lets double check with @Buwujiu.

surajgoraya (commented):
Small changes overall, I do have concerns about how we're doing the Rep signatures, but otherwise this looks good

apps/ui/src/api/mutations/useCreateSignature.ts
surajgoraya: Whoops sorry you're using the error handler! this is already, translated

surajgoraya (approved):

Issue #266, opened 2025-03-04
API Integration - Applicant Submit Revisions


Buwujiu:
## Descriptions
After the revisions API are done, integrate them with UI so that applicant can submit revisions on the portal. 
## Acceptance Criteria
- Applicant will receive revision notes via email, they can then log in to DACO portal to edit application. 
- After requests are sent, Only requested sections are open for edits, the rest should remain locked. 
- Applicant do not need to sign again if the signature already in the original application. 
- Once the application is submitted again, it will go to the next review state.

Buwujiu:
Working on a PR today.

Buwujiu:
found an issue in qa: I requested revisions as rep, then I login as applicant, but all the sections are open for edits, we want only the requested sections to open. The application I used for testing is https://daco-ui.pcgl-qa.cumulus.genomeinformatics.org/application/3/intro/

Buwujiu:
Tested in QA, working as expected! Closing!

PR #369, opened 2025-04-21
#266: Get Application Revisions & Applicant Submit Revisions API Integration

surajgoraya:
## Summary

This PR ended up becoming larger than expected, however all changes centre around the Application Revisions process. 

This PR integrates getting revisions for applications, unlocking the appropriate sections, and allowing the applicant to submit the revisions. In the process of integrating these endpoints to the UI however, various issues with the sate machine logic, as well the submit endpoint logic were fixed, alongside adding a DTO object, and moving the URL for the endpoint under `/revisions`.

### Related Issues
- #266 

## Important Note❗:

I've changed how `editApplication` works in the `applicationService`. It now no longer sets the State to `DRAFT` when edited. This seems like a holdover from when we didn't have the state manager. We shouldn't directly be touching the Application state any longer without using the state manager. I'm removing this, however if this causes issues, please advise.

### 🌸 UI
- Added a new hook `useSubmitRevisions`, is used to submit revisions 
- Added a new hook `useGetApplicationFeedback` which gets revisions
  - Both hooks are called on `pages/applications/index.tsx` when the page loads.
- Modified SectionFooter.tsx to add "Submit Revisions" actions
- Modified `SectionMenu.tsx` to display lock icon if new prop `isLocked` set to true (thanks for the help on this @JamesTLopez!)
- Added Lock Symbol to `SectionTitle`
- Added `revisions` and `state` to ApplicationOutletContext
- Modified `/sections` components
  - All components now use logic calculated in variable `canEdit` which takes into account if the section has revisions or `isInEditMode`
  - `sign.tsx` has some special logic which is used to enable the Submit Application button for when revisions exist.

### 🖥️  Server
- 🚥 **State Manager:** 
   - Updated to correct for a error in how state transitions worked.
      - Previously, the transition from `DAC/REP_REVIEW_REQUESTED` back to `DAC/REP_REVIEW` did not exist, however was required by our sate diagram
- Edit Application controller and service updated to stop auto transitioning every change to DRAFT mode.
- 🚏 **Application Router:** 
   - Various spelling mistakes corrected, fixed 404 returning INVALID_REQUEST instead of NOT_FOUND.
   - Moved `/:applicationId/submit-revision` to Revisions tag in swagger
       - 🔏 Added auth checks to the endpoint.
 - Added `convertToRevisionsRecord` function in `alias.ts` since route was previously returning snake_case objects.  

### 📦 Packages
- `/data-model` - Added `RevisionsDTO` object which concerted from snake_case to camelCase
- `/validation` - Corrected validation for the revisionDataSchema which returns null values.
   - Added Response Schama which extends from this.

### Special Instructions
Before running these changes, you will need to build `@pcgl-daco/data-model`:

```bash
# in the root of the project
pnpm build:all
```
## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/ui/src/api/queries/useGetApplicationFeedback.ts
```diff
@@ -0,0 +1,97 @@
...
+					project: {
+						isApproved: revisionDependant ? data[0]?.projectApproved : undefined,
+						comment: data[0]?.projectNotes ?? null,
+					},
+					study: {
+						isApproved: revisionDependant ? data[0]?.requestedStudiesApproved : undefined,
+						comment: data[0]?.requestedStudiesNotes ?? null,
+					},
+					ethics: {
+						isApproved: revisionDependant ? data[0]?.requestedStudiesApproved : undefined,
```
JamesTLopez, line 74: this should be `ethicsApproved`

apps/api/src/controllers/applicationController.ts
```diff
@@ -88,7 +102,7 @@ export const editApplication = async ({
 
 	if (!formattedResult.success) return formattedResult;
 
-	return await applicationRepo.editApplication({ id, update: formattedResult.data });
+	return await applicationRepo.editApplication({ id, update: formattedResult.data, shouldKeepState: shouldKeepState });
```
JamesTLopez, line 105: We can just pass `shouldKeepState`

```suggestion
	return await applicationRepo.editApplication({ id, update: formattedResult.data, shouldKeepState });
```

apps/api/src/controllers/applicationController.ts
```diff
@@ -518,7 +532,16 @@ export const getRevisions = async ({
...
 		}
 
-		return success(revisionsResult.data);
+		const aliasedRevs = revisionsResult.data.map((rev) => convertToRevisionsRecord(rev));
+
+		const filteredFailures = aliasedRevs.filter((results) => !results.success);
+		const filteredSuccesses = aliasedRevs.filter((results) => results.success).map((results) => results.data);
+
+		if (filteredFailures.length) {
+			throw new Error('Failed to alias Revisions Record');
```
JamesTLopez, line 541: Shouldn't we return the `failure("SYSTEM_ERROR",message)` pattern here?

apps/api/src/routes/applicationRouter.ts
```diff
@@ -86,7 +86,7 @@ async function validateUserPermissionForApplication({
 
 		// ensure application has this user's ID
 		if (applicationResult.data.userId !== userId) {
-			return failure('FORBIDDEN', 'User is not the creater of this application.');
```
JamesTLopez, line 89: 🐞 CREATER

apps/api/src/routes/applicationRouter.ts
```diff
@@ -922,21 +921,49 @@ applicationRouter.get(
...
 					});
 					return;
 				}
 
-				response.status(500).json({ error: result.error, message: result.message });
+				const result = await getRevisions({ applicationId: Number(applicationId) });
+
+				//Service only returns this if a SYSTEM_ERROR occurs, set to HTTP code 500 and bail if this is the case.
+				if (!result.success) {
+					response.status(500).json({ error: result.error, message: result.message });
```
JamesTLopez, line 956: Just curious, Is there a reason to do `.json` vs `.send` ? I don't think it matter either way

apps/ui/src/pages/applications/sections/sign.tsx
```diff
@@ -91,28 +104,36 @@ const SignAndSubmit = () => {
...
+			(Object.values(revisions).find((rev) => rev.isApproved === false) &&
+				state === 'INSTITUTIONAL_REP_REVISION_REQUESTED') ||
+			state === 'DAC_REVISIONS_REQUESTED';
+
+		if (hasRevisions || canEdit) {
+			return true;
+		} else if (role === 'APPLICANT') {
 			return data?.applicantSignature !== getValues('signature');
+		} else {
+			return data?.institutionalRepSignature !== getValues('signature');
```
JamesTLopez, line 127: oops thanks for catching this. Missed this condition

apps/api/src/controllers/applicationController.ts
surajgoraya: This will shortcut to the catch statement below which does return a failure

apps/api/src/routes/applicationRouter.ts
surajgoraya: Nothing! I think .json runs a serialization in-case there's any objects which need to be serialized before sending, but I think .send might do this too? 🤔

surajgoraya:
Tests are now passing, the previous tests had some incorrect assumptions on how the state machine was supposed to work.

image

apps/ui/src/pages/applications/sections/institutional.tsx
```diff
@@ -40,7 +40,9 @@ const rule = createSchemaFieldRule(institutionalRepSchema);
 
 const Institutional = () => {
 	const { t: translate } = useTranslation();
-	const { isEditMode } = useOutletContext<ApplicationOutletContext>();
+	const { isEditMode, revisions } = useOutletContext<ApplicationOutletContext>();
+	const canEdit =
+		(revisions.institutional?.isApproved !== undefined && !revisions.institutional?.isApproved) || isEditMode;
```
leoraba, line 45: I see this condition in every section, should we consider move it to a shared function to be more align with DRY principal?

packages/validation/src/routes/applicationRoutes.ts
```diff
@@ -103,25 +103,33 @@ export const basicApplicationResponseSchema = applicationResponseSchema.omit({ c
 
 export const revisionDataSchema = z
 	.object({
-		comments: z.string().optional(),
-		applicantNotes: z.string().optional(),
+		comments: z.string().optional().nullish(),
```
leoraba, line 106: looks like making this field `optional` and `nullish` is redundant. 
https://zod.dev/?id=nullish

apps/ui/src/pages/applications/sections/access.tsx
leoraba, line 1: not sure if this is related to this ticket. Are application `comments` and section `notes` displayed to the user?

apps/api/src/controllers/applicationController.ts
```diff
@@ -89,10 +90,21 @@ export const editApplication = async ({
...
 	const canEditResult = new ApplicationStateManager(application)._canPerformAction(edit);
 
+	/**
+	 * By default the application service will reset an applications state to DRAFT when edited. We need to avoid this
+	 * behaviour for any applications within the REP_REV or DAC_REV states.
+	 */
+	const shouldKeepState =
+		application.state === ApplicationStates.INSTITUTIONAL_REP_REVISION_REQUESTED ||
+		application.state === ApplicationStates.DAC_REVISIONS_REQUESTED;
+
```
leoraba, line 107: just looking through the code, looks like this is missing in the `fileController` and `signController` as well.

surajgoraya: After talking with Jon about this, the logic to auto set the state to DRAFT when edited needs actually just needs to be removed from the `editApplication` function in the service. It breaks the principal of relying on the State Machine and I think it was added before we had that working! Will refactor.

apps/ui/src/pages/applications/sections/access.tsx
surajgoraya: Not for MVP I believe :)

apps/api/src/controllers/applicationController.ts
surajgoraya: I've gone ahead and changed how `editApplication` works in the `applicationService`. It now no longer sets the State to `DRAFT` when edited. This seems like a holdover from when we didn't have the state manager. We shouldn't directly be touching the Application state any longer without using the state manager. I'm removing this, however if this causes issues, please advise.

apps/api/src/resources/swagger.yaml
```diff
@@ -988,6 +950,44 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ClientErrors'
+        '200':
+          description: 'Application successfully submitted for revision'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/responses/ApplicationResponse'
```
leoraba, line 989: Based on the `applicationRouter` code, this endpoint returns an object of type: 
```
{message: string; data: ApplicationRecord}
```

apps/ui/src/pages/applications/utils/canEditSection.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ *    If we're not in `DAC_REVISIONS_REQUESTED` or `INSTITUTIONAL_REP_REVISIONS_REQUESTED`
+ *    then the `isApproved` is always set to `undefined` in the API hook.
+ *    This is because we cannot edit outside of these modes (besides edit mode, more on that below),
+ *    and so we want to ensure that we always fail, even if there are technically "revisions"
+ *    from the previous cycle of revisions requests.
+ *
+ * 2. If we're currently in `/edit` mode, then we can always edit. `/edit` is only enabled when we're
+ *    in DRAFT mode.
+ *
```
JamesTLopez, line 41: Although I appreciate the description, this might be over-documenting. Could probably just remove the two points.

leoraba (approved):
LGTM 👍

JamesTLopez (approved):
Looks good!

Issue #374, opened 2025-04-24
Cannot Change Collaborator Position Title


Buwujiu:
## Descriptions
When editing collaborators, if the Position Title is pre-populated, changing it results in a server error. 
How to reproduce: https://app.screencast.com/p0OaK8zYOedGj
## Acceptance Criteria
- Position Title should be an optional field. 
- Should be able to update the field and save changes.

Buwujiu:
Tested in dev, working as expected!

PR #379, opened 2025-04-25
#168 & #374: ♻️ Refactor Collaborators API, fix deleting collaborators checks

surajgoraya:
## Summary

This PR refactors the Collaborators API by removing duplicate collaborators checks in the service and controller layer. It also makes other changes related to the API, now relying on the email in relation to the App ID as primary identifiers. This PR also adds a small utility to read Postgres Errors.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/168
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/374

## Description of Changes
This PR refactors the Collaborators API by removing duplicate collaborators checks in the service and controller layer. It instead now relies on the database with a composite primary key set on Application ID and Collaborator Institutional Email, which means that if a collaborator with the same institutional email is added again, the check will fail with a DB error. This allows us to conduct this check within our DB rather than relying on the service to do it.

This PR also changes various Service and Controller calls to rely on the collaborator email in relation to the App ID to be identifiers for updating and deleting rather than ID. 

### UI
- Updated collaborators hooks to use `CollaboratorInstitutionalEmail` and updated API paths.
- Updated `collaborators.tsx` to use `CollaboratorInstitutionalEmail` as key value.
- Removed "required" rule from edit modal position title
- Updated small copy issue

#### 🌸 **UX Improvements**
- Collaborator add and edit modals will re-highlight the email if you attempt to add a duplicate email.

### Server
- Updated `collaborators` table to drop ID, and use `application_id` and `institutional_email` as a composite primary key. Please see notes under special instructions.
- Removed checks from service and controller relating to duplicate records. This is now handled by the DB and returned by the service as a `DUPLICATE_RECORD` error if the DB error is hit.
- Converted all references to collaborator ID to be collaborator email instead.
   - Updated delete to specify in there where clause what the Application ID was, this fixed deleting issue Edmund found.  
- Updated data model and validation to reflect these changes
- Updated tests.

New: 
- Added `utils.ts` which has a small `isPostgresError` function along with `PostgresErrors` const, these can be used to determine when a error thrown in a service is due to a DB error. We use this here to determine if it's a key constraint violation. Please take a look at the notes above these functions for more info.

### Special Instructions
Before running these changes, you will need to build `@pcgl-daco/validation` and `@pcgl-daco/data-model` :
```bash
pnpm build:all
```
**Additionally**, you will need to run a migration.

> [!IMPORTANT]  
> This migration **WILL FAIL** if records exist in the DB that violate the primary key constraint (duplicate collaborators with the same institutional email belonging to the same application ID)
>
> Please take of these if necessary before running the migration.

```bash
cd /apps/api && pnpm drizzle-kit migrate
````

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### CODE REVIEW

**Comment by JamesTLopez** [comment]
@surajgoraya 

Testing on the UI, the create works great! But theres a weird UI/UX issue when we edit. 
If you try to edit the email address, and hit save, it will return successful but no changes were made to that email field.
I assume this is because we made a composite key from application_id and institutional_email, and postgres probably doesn't want anyone to alter the composite key. 

I think we should maybe disable that field when a user decides to edit a row, forcing them to delete and remake if they want a new email? Maybe this is a patrick question.

File: apps/api/src/controllers/collaboratorsController.ts [hunk 1]
```diff
@@ -242,20 +223,7 @@ export const updateCollaborator = async ({
...
-			collaboratorUpdates.collaboratorInstitutionalEmail === collaborator.institutional_email
-		);
-	});
-
-	if (hasDuplicateRecords) {
-		// TODO: List the duplicate records.
-		return failure('DUPLICATE_RECORD', `Cannot create duplicate collaborator records.`);
-	}
-
-	const { id } = collaboratorUpdates;
```
JamesTLopez, line 258: Mentioned this in a comment outside the review (if you have not seen it, please refer to that first), we may need to bring it back depending on the expected UI/UX flow, unless we can throw duplicate errors on the db level for update related CRUD

surajgoraya: We should be throwing duplicate errors at the DB level already given that updating a record to have the same Email as another ID should throw a similar constraint violation that add does!

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -196,18 +177,21 @@ export const listCollaborators = async ({
 /**
  * Update a selected collaborator by ID
  * @param application_id - ID of related application record to associate with Collaborators
+ * @param institutional_email - The institutional email of the collaborator you want to update
  * @param user_id - ID of Applicant updating the application
  * @param collaborators - Collaborator record with updated properties
```
leoraba, line 182: this param doesn't have the right name

apps/api/src/controllers/collaboratorsController.ts
```diff
@@ -242,35 +226,21 @@ export const updateCollaborator = async ({
...
-	}
-
-	const { id } = collaboratorUpdates;
-
 	const collaborator: Partial<CollaboratorModel> = {
 		first_name: collaboratorUpdates.collaboratorFirstName,
 		middle_name: collaboratorUpdates.collaboratorMiddleName,
 		last_name: collaboratorUpdates.collaboratorLastName,
 		suffix: collaboratorUpdates.collaboratorSuffix,
 		position_title: collaboratorUpdates.collaboratorPositionTitle,
```
leoraba, line 234: just curious, why is not the field `title` in here, and what's the difference with `position_title`

apps/api/src/service/collaboratorsService.ts
```diff
@@ -84,51 +57,86 @@ const collaboratorsSvc = (db: PostgresDb) => ({
...
+				);
 			}
 
 			return success(updatedRecord);
 		} catch (error) {
-			const message = `Error updating collaborator with id: ${id}`;
+			const postgresError = isPostgresError(error);
+
+			if (postgresError && postgresError.code === PostgresErrors.UNIQUE_KEY_VIOLATION) {
+				return failure('DUPLICATE_RECORD', `Cannot update record to be duplicate collaborator records.`);
```
leoraba, line 136: this texts looks weird. maybe `Cannot create duplicate collaborator records`

apps/api/tests/api/collaborators-controller.test.ts
```diff
@@ -180,9 +180,14 @@ describe('Collaborators Controller', () => {
...
+			const { institutional_email } = testCollaborators[0];
+
+			const collaboratorUpdates: CollaboratorDTO = {
+				collaboratorFirstName: 'Test',
+				collaboratorLastName: 'User',
+				collaboratorInstitutionalEmail: institutional_email,
+				collaboratorType: 'Test User',
+			};
 
 			const collaboratorResult = await updateCollaborator({
```
leoraba, line 192: running test fails. looks like this function is passing the id instead of the email

apps/api/src/service/applicationService.ts
```diff
@@ -194,19 +194,19 @@ const applicationSvc = (db: PostgresDb) => ({
...
 		try {
 			// TODO: These two queries can be combined into 1
 			// Fetch applicationId for the collaborator
 			const applicationIds = await db
 				.select({ application_id: collaborators.application_id })
 				.from(collaborators)
-				.where(eq(collaborators.id, collaboratorId));
+				.where(eq(collaborators.institutional_email, collaboratorEmail));
 			if (applicationIds.length === 0) {
 				return failure('NOT_FOUND', 'Could not find an collaborator for the provided collaboratorId');
```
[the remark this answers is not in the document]
surajgoraya: Good catch! Fixed. I don't actually believe this function is even being used anywhere, but I kept it around in-case it's for future functionality

apps/api/tests/api/collaborators-controller.test.ts
surajgoraya: Whoops! Good catch, updated
image

apps/api/src/controllers/collaboratorsController.ts
leoraba: thanks for clarifying 👍

leoraba (approved):
LGTM 🎉

Issue #357, opened 2025-04-15
API Integration - DAC Reject


Buwujiu:
## Descriptions
Integrate the reject endpoint so that when dac clicks the reject button, the application is rejected.
https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=2488-15890&t=oJg2ABQI8aTqadao-4
## Acceptance Criteria
- a modal should populate when clicking on the reject button: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=2488-16403&t=oJg2ABQI8aTqadao-4
- a modal should populate when reject is confirmed: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=2488-16941&t=oJg2ABQI8aTqadao-4
- App status should change to Rejected.

Buwujiu:
Tested in QA, working as expected! Closing!

PR #380, opened 2025-04-25
#357: Integrate application reject endpoint with modal confirmation

Azher2Ali:
#357 : Integrate application reject endpoint with modal confirmation

## Summary

Integrate the reject application workflow for DAC users. When the DAC clicks the "Reject" button on an application, a confirmation modal should appear. Upon confirming, the application's status is updated to "Rejected", a success modal is displayed, and the user is redirected to the manage applications page.

## Description of Changes

- Integrated the /applications/{applicationId}/reject endpoint into the DAC application management UI.
- Added a confirmation modal when the "Reject" button is clicked, as per [Figma design](https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=2488-16403&t=oJg2ABQI8aTqadao-4).
- Upon confirming rejection:
- The application's status is updated to "Rejected".
- A success modal appears, following [Figma design](https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=2488-16941&t=oJg2ABQI8aTqadao-4).
- The DAC user is redirected to the manage applications page.
- Updated API spec to reflect the use of applicationId as a path param for the reject endpoint.
- Performed backend response conversion to camelCase where required for frontend consistency.

### Special Instructions
```
pnpm i
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
  - The PR title is properly formatted to match the pattern: `#357:  Integrate application reject endpoint with modal confirmation`
- [x] **Labels Added**
  - Label is added for each package/app that is modified (`api`, `ui`)
- [x] **Local Testing**


apps/ui/src/api/mutations/useRejectApplication.ts
```diff
@@ -0,0 +1,63 @@
...
+			}).then(withErrorResponseHandler);
+
+			return await response.json();
+		},
+		onSuccess: () => {
+			notification.openNotification({
+				type: 'success',
+				message: translate('dacRejectSuccess'),
+			});
+			navigate('/dashboard');
```
surajgoraya, line 52: So following the pattern we started doing a few PRs ago regarding navigation, this navigate shouldn't be in the hook but rather in the `handleRejectApplication` button click handler.  Additionally, since there's a modal that needs to popup after we should only navigate after the user clicks OK on it:

Image showing the DAC Reject confirmation modal.


Also given that this modal is popping up we **do not** need to show a notification in this case.

apps/ui/src/components/pages/application/ApplicationViewerHeader.tsx
```diff
@@ -71,6 +74,13 @@ const ApplicationViewerHeader = ({ id, state }: AppHeaderProps) => {
 		});
 	};
 
+	const handleRejectApplicationRequest = () => {
+		rejectApplication({ applicationId: id }).then(() => {
+			setOpenRevisionsModal(false);
+			navigate('/dashboard');
+		});
+	};
```
surajgoraya, line 82: Oh, looks like you're already doing a navigate here, so yeah please remove it from the mutation hook. 

Additionally, we should show the success modal see: 

Image showing the DAC Reject confirmation modal.

I already created this modal for another piece of UI. You can reuse it. It's called `SuccessModal.tsx`. You will need to pass the appropriate success text via props.

apps/ui/src/api/mutations/useRejectApplication.ts
```diff
@@ -0,0 +1,63 @@
...
+		},
+		onError: () => {
+			notification.openNotification({
+				type: 'error',
+				message: translate('dacRejectFailed'),
```
surajgoraya, line 57: This reference in translate will fail and just display `dacRejectFailed`.  You need to reference the entire JSON path. You likely want `modals.rejectApplication.notificaitons.dacRejectFailed`

apps/ui/src/api/mutations/useRejectApplication.ts
```diff
@@ -0,0 +1,63 @@
...
+const useRejectApplication = () => {
+	const notification = useNotificationContext();
+	const { t: translate } = useTranslation();
+	const navigate = useNavigate();
+
+	return useMutation<ApplicationResponseData, ServerError, RejectApplicationPayload>({
+		mutationFn: async ({ applicationId }) => {
+			const response = await fetch('/applications/reject', {
+				method: 'POST',
+				body: JSON.stringify({ applicationId }),
```
surajgoraya, line 42: We should probably convert this API endpoint to just use our regular `/applications/:applicationId/{operation}` pattern removing the body so it just becomes `/applications/:applicationId/reject`.

We're not using the body for anything else than application ID now, and I think this PR would be a good place to change it over. You would need to change the router as well as the swagger file.

surajgoraya (dismissed):
Changes are required before this can be merged in.

We should probably convert this API endpoint to just use our regular `/applications/:applicationId/{operation}` pattern removing the body so it just becomes `/applications/:applicationId/reject`.

We're not using the body for anything else than application ID now, and I think this PR would be a good place to change it over. You would need to change the router as well as the swagger file.

apps/ui/src/components/pages/application/ApplicationViewerHeader.tsx
```diff
@@ -147,6 +157,20 @@ const ApplicationViewerHeader = ({ id, state }: AppHeaderProps) => {
...
+					width={'100%'}
+					style={{ top: '20%', maxWidth: '800px', paddingInline: 10 }}
+					open={showRejectModal}
+					onOk={handleRejectApplicationRequest}
+					onCancel={() => setShowRejectModal(false)}
+				>
+					<Flex style={{ height: '100%', marginTop: 20 }}>
+						<Text>{translate('modals.rejectApplication.description')}</Text>
+					</Flex>
+				</Modal>
```
JamesTLopez, line 173: Lets move this into its own component. Similar to how <SuccessModal .../> works. Pass down the useState values (showRejectModal and setShowRejectModal ) as props: 

```js
const [showRejectModal, setShowRejectModal] = useState(false);
```

Then take the logic for:
```ts
const handleRejectApplicationRequest = () => {
		rejectApplication({ applicationId: id }).then(() => {
			setOpenRevisionsModal(false);
			navigate('/dashboard');
		});
	};
```

and put it directly into the newly created modal component. This way we can keep our ApplicationViewerHeader component cleaner. We should probably move `CloseApplication` to another component as well, but just focus on this implementation.

JamesTLopez (changes requested):

apps/ui/src/api/mutations/useRejectApplication.ts
JamesTLopez: image

apps/ui/src/components/pages/application/ApplicationViewerHeader.tsx
JamesTLopez: Theres a folder `modals` in the same directory as `ApplicationViewerHeader` please add the new reject modal into there

apps/ui/src/api/mutations/useRejectApplication.ts
```diff
@@ -0,0 +1,61 @@
...
+				body: JSON.stringify({
+					applicationId: applicationId,
+					rejectionReason: rejectionReason,
+				}),
+			}).then(withErrorResponseHandler);
+
+			return await response.json();
+		},
+		onSuccess: async (data) => {
+			await queryClient.invalidateQueries({ queryKey: [`application-$`] });
```
JamesTLopez, line 50: Make sure to add the suggestions discussed offline first:

```suggestion
			await queryClient.invalidateQueries({ queryKey: [`application-${data.id}`] });
```

apps/api/src/routes/applicationRouter.ts
```diff
@@ -444,16 +444,16 @@ applicationRouter.post(
...
 
 applicationRouter.post(
-	'/reject',
+	'/:applicationId/reject',
 	authMiddleware({ requiredRoles: ['DAC_MEMBER'] }),
 	async (
 		request: Request,
 		response: ResponseWithData<
-			{ message: string; data: ApplicationRecord },
+			ResponseWithData<ApplicationResponseData>,
```
JamesTLopez, line 452:
```suggestion
		ApplicationResponseData,
```

JamesTLopez (approved):
🔥 LGTM!

Issue #348, opened 2025-04-09
Make the Entire Card Clickable


Buwujiu:
## Descriptions
Make the entire Start a new application card clickable, not just the Get started button. 

![Image](https://github.com/user-attachments/assets/b577f84d-8090-4975-938d-e6534454e61a)

Buwujiu:
Tested in dev, the entire card is clickable, however the Get Started button is gone. Please make sure the card aligns with the design. 

![Image](https://github.com/user-attachments/assets/74d41f71-17fd-4ce0-ab28-7e6b6293a803)

Buwujiu:
Design is changed to remove the Get Started button, so this is working as expected in dev! Thanks Azher, closing!

PR #383, opened 2025-04-28
#348: Make the entire "Start a New Application" card clickable

Azher2Ali:
#348 : Make the entire "Start a New Application" card clickable

## Summary

Made the entire "Start a New Application" card clickable, so that clicking anywhere on the card — not just the "Get Started" button — will trigger the application start action.

### UI
- Updated the ApplicationCard component to wrap the entire card in a clickable element (<Link> or onClick handler depending on implementation).
- Ensured both the card content and the "Get Started" button trigger the same navigation event.
- Updated relevant styling to show a pointer cursor on hover over the whole card.
- No new packages added.
-` ApplicationCard `handles both mouse and keyboard interactions, and follows the same ARIA-compliant pattern which have there with Enter and Space triggering the action.

### Special Instructions
```
pnpm i
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
  - The PR title is properly formatted to match the pattern: `#348: Make the entire "Start a New Application" card clickable `
- [x] **Labels Added**
  - Label is added for each package/app that is modified (`ui`)
- [x] **Local Testing**

### DIFF
--- a/apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
+++ b/apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
@@ -18,7 +18,7 @@
  */
 
 import useCreateApplication from '@/api/mutations/useCreateApplication';
-import { Button, Card, Flex, theme, Typography } from 'antd';
+import { Card, Flex, theme, Typography } from 'antd';
 import { useTranslation } from 'react-i18next';
 
 const { Title } = Typography;
@@ -30,28 +30,23 @@ const NewApplicationCard = () => {
 
 	const { mutate: createNewApplication } = useCreateApplication();
 
-	const handleCreateNewApplication = () => {
-		createNewApplication();
+	const handleCardClick = (event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>) => {
+		if (('key' in event && (event.key === 'Enter' || event.key === ' ')) || event.type === 'click') {
+			event.stopPropagation();
+			createNewApplication();
+		}
 	};
 
 	return (
 		<Card
 			style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200, cursor: 'pointer' }}
 			hoverable
-			onClick={handleCreateNewApplication}
+			tabIndex={0}
+			onClick={handleCardClick}
+			onKeyDown={handleCardClick}
 		>
 			<Flex justify="center" align="center" vertical gap="middle" style={{ height: '100%' }}>
 				<Title level={3}>{translate('dashboard.startNewApp')}</Title>
-				<Button
-					color="default"
-					variant="outlined"
-					onClick={(e) => {
-						e.stopPropagation();
-						handleCreateNewApplication();
-					}}
-				>
-					{translate('button.getStarted')}
-				</Button>
 			</Flex>
 		</Card>
 	);
--- a/apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
+++ b/apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
@@ -30,11 +30,26 @@ const NewApplicationCard = () => {
 
 	const { mutate: createNewApplication } = useCreateApplication();
 
+	const handleCreateNewApplication = () => {
+		createNewApplication();
+	};
+
 	return (
-		<Card style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200 }} hoverable>
-			<Flex justify="center" align="center" vertical gap="middle">
+		<Card
+			style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200, cursor: 'pointer' }}
+			hoverable
+			onClick={handleCreateNewApplication}
+		>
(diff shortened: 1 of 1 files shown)


apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
```diff
@@ -30,11 +30,26 @@ const NewApplicationCard = () => {
...
 				<Title level={3}>{translate('dashboard.startNewApp')}</Title>
-				<Button color="default" variant="outlined" onClick={() => createNewApplication()}>
+				<Button
+					color="default"
+					variant="outlined"
+					onClick={(e) => {
+						e.stopPropagation();
+						handleCreateNewApplication();
+					}}
+				>
```
surajgoraya, line 52: We likely want to get rid of the "Get Started" button entirely. So you likely want to remove completely, it's confusing for the User that we can click two buttons to do the thing. The card itself being clickable should be enough.

apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
```diff
@@ -30,11 +30,26 @@ const NewApplicationCard = () => {
...
 
+	const handleCreateNewApplication = () => {
+		createNewApplication();
+	};
+
 	return (
-		<Card style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200 }} hoverable>
-			<Flex justify="center" align="center" vertical gap="middle">
+		<Card
+			style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200, cursor: 'pointer' }}
```
surajgoraya, line 39: Good shout on making the cursor a pointer! that's good UX!

apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
```diff
@@ -30,11 +30,26 @@ const NewApplicationCard = () => {
 
 	const { mutate: createNewApplication } = useCreateApplication();
 
+	const handleCreateNewApplication = () => {
+		createNewApplication();
+	};
+
```
surajgoraya, line 36: This is small and nitpicky but, if you look at the `handleCardClick` event handler inside ApplicationCard.tsx, we likely want to do something similar and enable the use of keyboard controls on the card itself. 

```tsx
// ... ApplicaitonCard.tsx

const handleCardClick = (event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>) => {
		/**
		 * For keyboard navigation we need to make sure we abide by AIRA best practices. Check for the
		 * enter key or the space bar.
		 *
		 * See for more info: https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/
		 */
		if (('key' in event && (event.key === 'Enter' || event.key === ' ')) || event.type === 'click') {
			event.stopPropagation();
			//....
		}
	};

```

Take a look at the `ApplicationCard.tsx` file for how we handle stuff there.

surajgoraya (changes requested):

apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
Azher2Ali: I've removed the "Get Started" button , but the card looks way to empty now. Should be approach @patrickdsantos for further suggestions?

apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
Azher2Ali: Thanks!

apps/ui/src/components/pages/dashboard/cards/NewApplicationCard.tsx
```diff
@@ -30,13 +30,23 @@ const NewApplicationCard = () => {
...
+			event.stopPropagation();
+			createNewApplication();
+		}
+	};
+
 	return (
-		<Card style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200 }} hoverable>
-			<Flex justify="center" align="center" vertical gap="middle">
+		<Card
+			style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200, cursor: 'pointer' }}
```
surajgoraya, line 42: One last thing Azher - could we centre this?

image


You'll likely need to use Ant Design's `styles` prop and set the body of the card to take up 100%:
```suggestion
		<Card
		  styles={{
				body: {
					height: '100%',
				},
			}}
			style={{ backgroundColor: token.colorWhite, minHeight: 200, height: 200, cursor: 'pointer' }}
```

surajgoraya (changes requested):
One last thing! Sorry Azher. We should probably fully centre the card text:

image

Azher2Ali:
> One last thing! Sorry Azher. We should probably fully centre the card text:
> 
> image

I had this change in my mind, thanks for confirming!

surajgoraya (approved):
ship it! :shipit:

patrickdsantos:
Looks good! Thanks @surajgoraya for the suggestion. @Azher2Ali I don't think it needs any more design now that the text is centred. Thanks team! :)

Issue #336, opened 2025-04-01
Add a Mechanism to Redirect Reps


Buwujiu:
## Descriptions
After the reps login using CILogon, we need a process or mechanism to redirect them to the application that they are about to review.
## Acceptance Criteria
- After reps login, they should see the Introduction page of the application. see mock: 

![Image](https://github.com/user-attachments/assets/49c6f1ea-2986-46f7-bc41-c397bd0c1455)

- There won't be a My Applications page for them. 
- They should only see the application viewer. 
- connect the Login button with CILogon.

Issue #321, opened 2025-03-27
Apply for Access Button Should Populate Modal


Buwujiu:
## Descriptions
The Apply For Access button on the homepage should populate a modal: 

![Image](https://github.com/user-attachments/assets/908a8bab-9b76-4f20-91a9-d4a21e7a1f62)

![Image](https://github.com/user-attachments/assets/759edefa-251a-4fd8-8ab1-dfd7ce2aedaa)

Buwujiu:
Tested in dev, working as expected! Thanks!

PR #391, opened 2025-05-05
#336: Add Mechanism to Redirect Reps and Improve Login UX

surajgoraya:
## Summary

Adds a new page where Reps will be expected to Login to review an application. It also introduces a few changes to how logins redirect, by introducing a redirection mechanism on the frontend based on role. 

Finally, it also enhances the UX during user fetching by adding a Fullscreen Loader component which will appear during the "logging in" stage or redirecting stage. This helps reduce the user seeing the UI jump around, open the suggestions in-case this should be changed.

### Related Issues
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/336
- #321 

### UI
- Added a new page `/login/redirect` which will be the new destination for the API to redirect to, once a login is completed.
  - This component will, based on the user role redirect them to the dashboard, the manage applications page, or a specific application they're meant to review (more on that below).
  - It will also display a loader while this process happens to improve the login UX.  
- ✨ **New:**  Added a Local Storage utility file which will be used to clear, get, and set additional session information for the User.
   - Currently, this is only used for REP users, who after landing on `/review/:applicationId` have the Application ID stored in local storage using these functions. This is then used by the `/login/redirect` page to redirect Reps to the appropriate page.
- ✨ **New:** Added a new component `FullscreenLoader` which can display a full screen loading circle with text accompanying it. It can also display an overlay covering all of the UI Chrome if needed via a prop.
- ✨ **New:** Added `/pages/review` - This is the Rep review page. It can be accessed by URL `/review/:applicationId`
- ✨ **New:** Generalized Apply For Access modal used on the Get Started button on the home page and in the Menu Bar
  - This addresses #321 

### Server
- Made `authConfig.ts` use the new redirect path
- Fixed/Added auth role protections to `/applications` and `applications/revisions`
- AuthRouter now uses the service call `getUserRole` instead of it's own var.

### Validation
- Added validation for local storage objects

### Special Instructions
Before running these changes, you will need to build `@pcgl-daco/validation`:
```
pnpm build:all
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/ui/src/pages/login/redirect/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+	const { t: translate } = useTranslation();
+	const { role } = useUserContext();
+	const navigate = useNavigate();
+
+	const redirectRep = () => {
+		const sessionInfo = getExtraSessionInformation();
+
+		if (!sessionInfo) {
+			return '/';
+		}
```
JamesTLopez, line 37: We can just have this as the default return value and we can remove ` ?? '/'` from the `INSTITUTIONAL_REP ` case in the useeffect

apps/ui/src/pages/review/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+				...landingPageOuter,
+			}}
+		>
+			<Flex justify="center" align="center" vertical flex={1} style={{ ...landingPageInner }}>
+				<Flex flex={1} justify="center" vertical gap={'3rem'}>
+					<Flex align="flex-start" justify="center" vertical gap={'1rem'}>
+						<Title level={1}>{translate('repView.title')}</Title>
+						<Text style={landingPageDescription}>{translate('repView.description')}</Text>
+					</Flex>
+					<div>
```
JamesTLopez, line 116: Just curious, why `div`?

apps/ui/src/pages/review/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+		margin: token.marginLG,
+		padding: token.marginLG,
+	};
+
+	const landingPageDescription: React.CSSProperties = {
+		maxWidth: '35rem',
+	};
+
+	useEffect(() => {
+		if (isLoggedIn && loggedInRole !== 'INSTITUTIONAL_REP') {
```
JamesTLopez, line 61: Should this be just `!isLoggedIn || loggedInRole !== 'INSTITUTIONAL_REP'`? 
If the user is not logged in OR not a rep then we redirect? 

Could be wrong

apps/ui/src/components/Header.tsx
```diff
@@ -80,6 +88,44 @@ const HeaderComponent = () => {
...
+				target: '_self',
+			};
+		} else if (isLoggedIn && role === 'DAC_MEMBER') {
+			return {
+				name: translate('links.manageApplications'),
+				href: '/manage/applications',
+				position: 'right',
+				target: '_self',
+			};
+		} else if (isLoggedIn && role === 'APPLICANT') {
```
JamesTLopez, line 117: I think isLoggedIn is redundant here from the first condition.

apps/ui/src/components/FullscreenLoader.tsx
```diff
@@ -0,0 +1,48 @@
...
+		console.error(
+			'Error: FullscreenLoader - Using trueFullscreen with loadingText provided will result in text appearing behind the loader. ',
+		);
+	}
+
+	return (
+		<Content style={{ display: 'flex' }}>
+			<Flex align="center" justify="center" vertical gap={'2rem'} flex={1}>
+				<Spin fullscreen={trueFullscreen} indicator={<LoadingOutlined style={{ fontSize: 56 }} spin />} />
+				{loadingText ? <Text>{loadingText}</Text> : null}
```
JamesTLopez, line 42: I don't think its a bad idea to add default translated loading text if `loadingText` is not provided. This is optional tho, this works!

apps/ui/src/pages/review/index.tsx
surajgoraya: This prevented the button from going full width; without it the Flex causes the button to take the full width. I can use a `Flex` comp to wrap it too, but I thought `div` would be ok:
image

apps/ui/src/pages/review/index.tsx
surajgoraya: No, in this case we want to check if the user is Logged in and the role is not REP, if that's the case I want to send them to the login redirector which will send them to the appropriate dashboard (manage or regular). 

This is if a DAC Reviewer or a Applicant lands on the Rep Review login page.

JamesTLopez (approved):
🔥

apps/ui/src/global/localStorage.ts
```diff
@@ -0,0 +1,54 @@
...
+	const validSessionInfo = ExtraSessionInformationSchema.safeParse(sessionInfo);
+
+	if (validSessionInfo.success) {
+		localStorage.setItem(SESSION_INFO_KEY, JSON.stringify(sessionInfo));
+	}
+
+	return validSessionInfo.success;
+}
+
+export { clearExtraSessionInformation, getExtraSessionInformation, setExtraSessionInformation };
```
joneubank, line 54: Why not just export the functions directly?

apps/ui/src/global/localStorage.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { ExtraSessionInformation, ExtraSessionInformationSchema } from '@pcgl-daco/validation';
+
+const SESSION_INFO_KEY = 'pcgl-daco-session';
+
+function clearExtraSessionInformation() {
+	localStorage.removeItem(SESSION_INFO_KEY);
+}
+
+function getExtraSessionInformation() {
```
joneubank, line 28: It may be good here to declare the return type explicitly here so that at a glance we know the data type we'll get back. This will help devs reading the function to know what to expect and ensure that all paths in the function return one of the expected types.

apps/api/src/config/authConfig.ts
```diff
@@ -49,7 +49,7 @@ function getAuthConfig() {
 		throw new EnvironmentConfigError(`db`, parseResult.error);
 	}
 
-	return { ...parseResult.data, enabled, loginRedirectPath: '/dashboard', logoutRedirectPath: '/' };
+	return { ...parseResult.data, enabled, loginRedirectPath: '/login/redirect', logoutRedirectPath: '/' };
```
joneubank, line 52: 👍

apps/api/src/routes/applicationRouter.ts
```diff
@@ -1024,7 +1024,7 @@ applicationRouter.post(
 
 applicationRouter.get(
 	'/:applicationId/revisions',
-	authMiddleware({ requiredRoles: ['APPLICANT', 'DAC_MEMBER'] }),
+	authMiddleware({ requiredRoles: ['APPLICANT', 'INSTITUTIONAL_REP', 'DAC_MEMBER'] }),
```
joneubank, line 1027: Pretty sure that you can skip `requiredRoles` when it is all roles (other than anonymous).

apps/ui/src/components/FullscreenLoader.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const { Content } = Layout;
+const { Text } = Typography;
+
+interface FullscreenLoaderProps {
+	trueFullscreen?: boolean;
+	loadingText?: string;
+}
+
+const FullscreenLoader = ({ trueFullscreen = false, loadingText }: FullscreenLoaderProps) => {
```
joneubank, line 31: We need some comments (ts docs please) to explain to future people what trueFullscreen and loadingText do.

apps/ui/src/pages/dashboard.tsx
```diff
@@ -53,6 +59,16 @@ const DashboardPage = () => {
...
 
+	/**
+	 * Will auto redirect DAC Members and Institutional Reps to
+	 * their correct routes on component mount.
+	 */
+	useEffect(() => {
+		if (userInfo.isLoggedIn && userInfo.role !== 'APPLICANT') {
+			navigate('/login/redirect', { replace: true });
+		}
+	}, [navigate, userInfo.isLoggedIn, userInfo.role]);
```
joneubank, line 70: Shouldn't this be handled in a router file? this kind of logic hidden in the page itself may be hard to trackdown in the future.

apps/ui/src/pages/review/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+			navigation('/login/redirect', { replace: true });
+		}
```
joneubank, line 63: Repeat of previous comment... Can't remember what we decided for where this redirect logic should live, but it feels like it should be in the router component not in a page's use effect.

apps/ui/src/pages/review/index.tsx
joneubank: Out of curiosity, what tags does the Flex component add to the DOM?

packages/validation/src/localStorage/sessionExtras.ts
```diff
@@ -0,0 +1,38 @@
...
+		role: z.literal(userRoleSchema.Values.APPLICANT),
+	}),
+	z.object({
+		role: z.literal(userRoleSchema.Values.DAC_MEMBER),
+	}),
+	z.object({
+		role: z.literal(userRoleSchema.Values.ANONYMOUS),
+	}),
+]);
+export type ExtraSessionInformation = z.infer<typeof ExtraSessionInformationSchema>;
```
joneubank, line 38: Love me a good discriminated union!

apps/ui/src/pages/review/index.tsx
surajgoraya: image

`Flex` adds in a `div` as well, however, it pre-applies the `antd` classes

Issue #397, opened 2025-05-08
Remove Buttons For Closed App View


Buwujiu:
## Descriptions
steps to reproduce: 
1. Close Rep Review app as an applicant. 
2. Go to the app viewer, the REquest Revisions and Reject buttons are available for users.
Expected result: 
After closing app, dac or applicant should not have permission to request revisions or reject app anymore.
## Acceptance Criteria
Remove the two buttons on closed apps:

![Image](https://github.com/user-attachments/assets/ac20640c-1377-4d9b-94e7-45ffbeb242ee)

Buwujiu:
Tested in dev, buttons are gone as expected! Thanks!

Issue #378, opened 2025-04-24
Add Revoke Application Button


Buwujiu:
## Descriptions
Add a Revoke button to the application viewer. 
Here is how the UI works:

For applicant, application approved -> wants to revoke -> go to application viewer -> not allowed to edit anything -> UI shows that the current state is Approved either through the progress bar or something else -> applicant can use a Revoke button to revoke app -> modal pops up and asks for confirmation -> applicants confirms -> app status changes to "Revoked" -> app is still read-only but no action can be performed on it anymore.
For DAC, application approved -> wants to revoke -> go to Manage application page -> not allowed to edit anything -> UI shows that the current state is Approved either through the progress bar or something else -> applicant can use a Revoke button to revoke app -> modal pops up and asks for confirmation -> applicants confirms -> app status changes to "Revoked" -> app is still read-only but no action can be performed on it anymore.

Mocks: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=4450-14759&t=iiVg1XWAx0eUYMhO-4
https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=5089-16445&t=iiVg1XWAx0eUYMhO-4
https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=5089-16931&t=iiVg1XWAx0eUYMhO-4

## Acceptance Criteria
- [ ] add a revoke button and modal
- [ ] show the approved status in viewer
- [ ] show the revoked status in viewer

Buwujiu:
Tested in QA, working as expected, great work @surajgoraya !

Issue #359, opened 2025-04-15
Integration Revoke Application


Buwujiu:
## Descriptions
Integrate the revoke app endpoint (https://github.com/Pan-Canadian-Genome-Library/daco/issues/123) with the Revoke Button. 
Can start once UI button is ready. 
## Acceptance Criteria

- [ ] Apps can only be revoked after they are approved.
- [ ] Only applicant and DAC can revoke. 
- [ ] No action can be performed on revoked apps.

PR #400, opened 2025-05-12
#359, #378 & #397: Revoke Application Integration, Application Header Fixes & Role-Based Component Protection

surajgoraya:
## Summary
This PR integrates the Revoke Application API and introduces the required modals and buttons. It also introduces the ability for UI elements to be hidden according to user Role and or Application State. Finally, it makes some fixes to the Application Header, moving code which should be in its own components, as well as some bugs in the Application Router.

### Related Issues

- #359
- #378 
- #397

### UI
- **✨ New:** 
  - Added `RevokeApplication.ts` - API hook to revoke applications
  - Added Revoke application button and integrated API
  - `ProtectedComponent.tsx` - New component 
    -  Which, when wrapped over another component will only make it appear only when the user is a certain role and/or when the application is within a certain state
    - This component functions a lot like the ProtectedRoute component.

- **🔧 Fixes:**
  - `CloseApplicationModal.tsx` - Moved close application modal out of the header and into its own file.
  - Added `applicationState` into App Context, this is used by the the `ProtectedComponent` component.
  - Added TSDoc comments to `FullscreenLoader.tsx`
  - Made `localStorage.ts` cleaner, added TS return values.
  - Removed `useEffect` calls in the review page
  - Moved Rep Revision logic into its modal instead of in the App Header.
  - Made buttons in app header appear dynamically based off role and app state via the `ProtectedComponent` component. 
 
- **🌸 Refactors:**
  - Refactor Application Status Steps, to not use Enums, and rely on the translations file
  - Aligned UI closer to mockups
  - Made Revoked, Closed and Rejected states appear

### Server
- Added types to Request vars so TS would stop complaining.
- Fixed permissions on close endpoint
- Fixed access issues for DACs and Reps for app details.

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] Documentation

### DIFF
--- a/apps/ui/src/api/mutations/useRevokeApplication.ts
+++ b/apps/ui/src/api/mutations/useRevokeApplication.ts
@@ -25,13 +25,13 @@ import { type ApplicationDTO } from '@pcgl-daco/data-model';
 import { useMutation } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
 
-const useWithdrawApplication = () => {
+const useRevokeApplication = () => {
 	const notification = useNotificationContext();
 	const { t: translate } = useTranslation();
 
 	return useMutation<ApplicationDTO, ServerError, { applicationId?: string | number }>({
 		mutationFn: async ({ applicationId }) => {
-			const response = await fetch(`/applications/${applicationId}/withdraw`, {
+			const response = await fetch(`/applications/${applicationId}/revoke`, {
 				method: 'POST',
 			}).then(withErrorResponseHandler);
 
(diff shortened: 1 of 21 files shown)


apps/ui/src/api/mutations/useRevokeApplication.ts
```diff
@@ -0,0 +1,63 @@
...
+	const notification = useNotificationContext();
+	const { t: translate } = useTranslation();
+
+	return useMutation<ApplicationDTO, ServerError, { applicationId?: string | number }>({
+		mutationFn: async ({ applicationId }) => {
+			const response = await fetch(`/applications/${applicationId}/revoke`, {
+				method: 'POST',
+			}).then(withErrorResponseHandler);
+
+			return await response.json();
```
joneubank, line 38: This returns type `any`, we should validate the content we get back. If this is how they all work then maybe we don't start updating this now, but this is risky if the endpoint changes the return type and the ui hook doesn't keep up to date.

apps/ui/src/components/pages/application/ApplicationViewerHeader.tsx
```diff
@@ -47,58 +45,25 @@ type AppHeaderProps = {
...
 	const [showEditModal, setShowEditModal] = useState(false);
-	const { mutateAsync: repRevision } = useRepRevisions();
-	const notification = useNotificationContext();
-	const { mutateAsync: closeApplication, isPending: isClosing } = useCloseApplication();
 	const [showRejectModal, setShowRejectModal] = useState(false);
-	const [showSuccessRejectsModal, setShowSuccessRejectsModal] = useState(false);
-	const [showApproveModal, setShowApproveModal] = useState(false);
+	const [showRevokeModal, setShowRevokeModal] = useState(false);
+	const [showApprovalModal, setShowApprovalModal] = useState(false);
+	const [showRejectSuccessModal, setShowRejectSuccessModal] = useState(false);
```
joneubank, line 60: What happens if we show multiple modals at the same time?

joneubank (dismissed):

apps/ui/src/components/pages/application/ApplicationViewerHeader.tsx
surajgoraya: Hmm, theoretically, this is possible, however I don't think there ever is a scenario where multiple modals can ever open. If you click on the outside of the modal (in an attempt to click the buttons), it will auto dismiss the modal, and set the state to false.

JamesTLopez (approved):

JamesTLopez (approved):

Issue #425, opened 2025-06-02
Application is Editable to DAC


Buwujiu:
**Describe the bug**
When the application is in Rep Revision, it becomes editable to DAC. 
**To Reproduce**
1. Go to https://daco-ui.pcgl-qa.cumulus.genomeinformatics.org/, login as DAC
2. Make sure there is an app in rep revisions status
3. Click on the rep revisions app
4. The app is editable

**Expected behavior**
Applications that are under Rep Revisions should be read-only for DACs.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Buwujiu:
Tested in QA, working as expected! Closing!

PR #427, opened 2025-06-10
#425: Resolve Application Editable by DAC Users

JamesTLopez:
## Summary

Resolve users outside of APPLICANT to be able to make changes to the application.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/425

### API

- Add APPLICANT role to authmiddle where for `/edit` post endpoint

### UI

- Add additional logic to check the user role as well as the application-state lifecycle for page redirect
   - If the application-state is not DRAFT or if the user is not an APPLICANT, push back to view mode
- Add additional logic to check the user role on sidemenu icon selector
- Add additional logic to disable revision logic depending on user role
   - if the user role is NOT applicant, then the sidemenu will display a lock for all sections (minus sign for reps) 
   - if the user role is NOT applicant, pass empty revision object in Outlet Context

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### DIFF
--- a/apps/ui/src/components/pages/application/SectionMenu.tsx
+++ b/apps/ui/src/components/pages/application/SectionMenu.tsx
@@ -27,6 +27,7 @@ import SectionMenuItem from '@/components/pages/application/SectionMenuItem';
 import { VerifyFormSections, VerifySectionsTouched } from '@/components/pages/application/utils/validators';
 import { ApplicationSectionRoutes } from '@/pages/AppRouter';
 import { useApplicationContext } from '@/providers/context/application/ApplicationContext';
+import { useUserContext } from '@/providers/UserProvider';
 import { ApplicationStateValues } from '@pcgl-daco/data-model';
 import { SectionRevision, SectionRoutes, SectionRoutesValues } from '@pcgl-daco/validation';
 import { ValidateAllSections } from './utils/validatorFunctions';
@@ -44,6 +45,7 @@ const SectionMenu = ({ currentSection, isEditMode, appId, revisions, appState }:
 	const { state } = useApplicationContext();
 	const { mutate: editApplication } = useEditApplication();
 	const { data, isLoading } = useGetCollaborators(appId);
+	const { role } = useUserContext();
 
 	const handleNavigation: MenuProps['onClick'] = (e) => {
 		if (state?.formState.isDirty) {
@@ -91,7 +93,7 @@ const SectionMenu = ({ currentSection, isEditMode, appId, revisions, appState }:
 			selectedKeys={[currentSection]}
 			mode="inline"
 			items={
-				!isLoading
+				!isLoading && role
 					? ApplicationSectionRoutes.map((item) => {
 							const route = item.route;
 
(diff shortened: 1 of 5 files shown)


apps/ui/src/pages/applications/index.tsx
```diff
@@ -56,13 +59,16 @@ const ApplicationViewer = () => {
 
 	useEffect(() => {
 		if (applicationData && !applicationError) {
-			const isNotInEditLifecycle = applicationData.state !== ApplicationStates.DRAFT;
+			// Application can only be in edit if the application-state is in DRAFT and if the user is an APPLICANT
+			// TODO: possibly need to change depending on the auth rework with authz
+			const forceToViewMode =
+				applicationData.state !== ApplicationStates.DRAFT || role !== userRoleSchema.Values.APPLICANT;
```
leoraba, line 65: thanks for adding the comments, maybe just a suggestion to reverse the code logic to make it easier to read and matching with the comment. example:
```
if (!isEditMode) return;

const canEditApplication = 
    applicationData.state === ApplicationStates.DRAFT && 
    role === userRoleSchema.Values.APPLICANT

if(!canEditApplication){
// redirect to read-only page
}
```

apps/ui/src/components/pages/application/SectionMenu.tsx
```diff
@@ -91,7 +93,7 @@ const SectionMenu = ({ currentSection, isEditMode, appId, revisions, appState }:
 			selectedKeys={[currentSection]}
 			mode="inline"
 			items={
-				!isLoading
+				!isLoading && role
```
joneubank, line 96: Isn't `ANONYMOUS` a role, what does this check accomplish? Is it potentially undefined? Do we need an `isLoading` for user context?

joneubank (approved):
LGTM

apps/ui/src/components/pages/application/SectionMenuItem.tsx
```diff
@@ -34,6 +35,7 @@ export type SectionMenuItemProps = {
 	isLocked?: boolean;
 	hasCollaborators?: boolean;
 	appState: ApplicationStateValues;
+	role?: UserRole;
```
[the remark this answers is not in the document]
JamesTLopez: We do, the `role` can determine which Icon is being displayed on the side-navigation. Originally, what was displayed on the side-navigation was determined only by the current state of the app...`DRAFT | DAC_REVIEW | DAC_REQUEST_REVISIONS etc`. 

If the state is revisions requested, the sections that revisions were needed are unlocked and have no icon. The problem was, it showed no icon for ALL roles where it should ONLY be the `APPLICANT` role have the sections unlocked.

Heres the logic on `components/pages/application/utils/IconSelector.tsx`:

```js
// This is the logic for the application state when revisions are requested for both INSTITUTIONAL REP and DAC MEMBERS
const RevisionsRequested = ({ label, isLocked, role }: Omit<SectionMenuItemProps, 'appState'>) => {
	if (label === SectionRoutes.INTRO) {
		// do not display intro icon
		return <LockOutlined />;
	} else if (!isLocked && role === userRoleSchema.Values.APPLICANT) {
		// Do not display icon on sections with revisions AND if the user is APPLICANT.
		return;
	} else if (label === SectionRoutes.SIGN) {
		return;
	}
	return <LockOutlined />;
};
```

Issue #132, opened 2024-12-19
Manage Application - add search by app ID bar


Buwujiu:

edsu7:
Tested in QA. LGTM

edsu7:
Noticed columns sorts are not working

https://github.com/user-attachments/assets/cee4514b-d925-423b-92b3-7221184bce6b

PR #485, opened 2025-10-21
#475: Manage Applications Search Bar Implemenation

JamesTLopez:
## Summary

Full-stack implementation of search bar feature in the manage application page. Allows DAC users to be able to search for specific applications based of IDs, first and last names, institution names, and email addresses.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/132

### UI
- Added Search Bar UI
- Remove `useGetApplicationCounts` react-query hook
- Added additional response object to `useGetApplicationList` hook
   - now includes `totals`, which replaces api response from `useGetApplicationCounts`
- Added translations
   - Added translation for Search placeholder
   - Added translations for `Search Results for {{value}}`
- Refactor Manage page
   - Move helper functions to another file ... `apps/ui/src/components/pages/manage/utils/manageUtils.ts`
- Resolve Manage page bugs
   - Switching filtering options / pagination and search were cause double api calls. Resolved by limited useEffect calls dependency array
   - Search bar was re-rendering on refetch, resolve by moving it out of the re-rendering component
 
 
### Server
- REMOVE `application/metadata/counts` endpoint
- ADD`totals` response to `/applications` get all Applications endpoint
   - Now will have an additional response to return all the total counts  of each applications state
   - Replaces `application/metadata/counts` endpoint
- ADD weighted `search_index` to migrations for table `application_contents`
   - ADD `transformSearchIntoQuery` prevent repeat search sql
   - ADD `customCount` to properly query for state totals 
- REMOVE `application/metadata/counts` tests cases
- [x] <s>ADD `totals` test cases </s>
    - [x] Theres a bug with the test cases for `ListApplications`.  Created a separate PR https://github.com/Pan-Canadian-Genome-Library/daco/pull/492
   

### Package
- Relocate `ApplicationStateTotals` to shared package

### Special Instructions
Before running these changes, you will need to rebuild/install/migrate because of changes in shared packages and db:
```
pnpm drizzle-kit migrate
pnpm i && pnpm build:all
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### DIFF
--- a/apps/ui/src/components/pages/manage/DashboardFilter.tsx
+++ b/apps/ui/src/components/pages/manage/DashboardFilter.tsx
@@ -73,12 +73,9 @@ const DashboardFilter = ({ onFilterChange, filters, availableStates }: Dashboard
 			: filterStates.filter((t) => t !== selectedFilter);
 
 		/**
-		 * If the user selects total, it makes no sense to have any other filters selected,
-		 * in this case, we empty out the selected filters, adn replace them with just TOTAL
-		 *
-		 * Also if the user unselects all filters, default back to total.
+		 * If the user selects total, replace them with just TOTAL or if the user unselects all filters, default back to TOTAL.
 		 **/
-		if (selectedFilter === 'TOTAL' || (nextSelectedFilters.length === 0 && filters.length === 1)) {
+		if (selectedFilter === 'TOTAL' || nextSelectedFilters.length === 0) {
 			nextSelectedFilters = ['TOTAL'];
 		} else if (filterStates.includes('TOTAL')) {
 			/**


apps/api/src/service/applicationService.ts
```diff
@@ -267,25 +266,39 @@ const applicationSvc = (db: PostgresDb) => ({
...
+				const sanitizedSearch = searchText?.trim().replace(/\s+/g, ' | '); // add OR between phrases
+
+				let searchQuery;
+				// Use ILIKE for better partial matching
+				const search = `%${sanitizedSearch}%`;
+				searchQuery = sql`(
+						${applicationContents.application_id}::text ILIKE ${search} OR
+						(COALESCE(${applicationContents.applicant_first_name}, '') || ' ' || COALESCE(${applicationContents.applicant_last_name}, '')) ILIKE ${search} OR
+						COALESCE(${applicationContents.applicant_institutional_email}, '') ILIKE ${search} OR
+						COALESCE(${applicationContents.applicant_primary_affiliation}, '') ILIKE ${search}
```
leoraba: curious to see this change, `ILIKE` search is less efficient and won't use the GIN index. I'll be inclined to use the search like Drizzle [documentation](https://orm.drizzle.team/docs/guides/postgresql-full-text-search) if possible.

JamesTLopez: I agree that its definitely less efficient, but for the scope of our application it shouldn't be too much of an issue.
I reverted it back to the solution Drizzle suggests. Lets keep this solution in mind if the testers want potentially more lenient matching solution.

apps/ui/src/components/pages/manage/utils/manageUtils.ts
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Processes a page number passed into it, checking if it's valid, and if it needs to be converted into a int.
+ *
+ * Additionally, since page numbers on the server count from 0, whereas `antd` counts from 1, it can convert from
+ * either or using the `forAPI` param.
+ * @param pageNumber The current page that the user is on in the UI from the URL
+ * @param forAPI If the processed page number you'd like back is being sent to the server (count from zero)
+ * @returns A `number` that has been validated and converted.
+ */
+export const parsePageNumber = (pageNumber?: number | string | null, forAPI?: boolean): number => {
```
joneubank: The `forAPI` parameter has a bad code smell... it causes this function to have a mix of responsibilities... this function should be determining the page number in a consistent way (for instance, zero indexed), and it is the responsibility of the consuming code to know if it needs to modify that value for a display purpose or for communication with an API. Making this function need to know where the value will be used and modify the result accordingly means this function needs to mix in logic that is not part of its Single Responsibility (SOLID principles) of finding the page. Now we get a function with a bunch of confusing ternaries that modify the value up or down by one, but the consuming code is still responsible for making sure it has the correct value anyways. This is a recipe for off by one errors.

apps/ui/src/components/pages/manage/DashboardFilter.tsx
```diff
@@ -78,7 +78,7 @@ const DashboardFilter = ({ onFilterChange, filters, availableStates }: Dashboard
 		 *
 		 * Also if the user unselects all filters, default back to total.
 		 **/
-		if (selectedFilter === 'TOTAL' || (nextSelectedFilters.length === 0 && filters.length === 0)) {
+		if (selectedFilter === 'TOTAL' || (nextSelectedFilters.length === 0 && filters.length === 1)) {
```
joneubank: Is this correct? We only do this default when `filters.length === 1`?  I don't know the meaning of these checks, it just seems like an unusually specific case. I would've expected `filters.length > 0` or `filters.length === 0`, but specifically only doing this with one filter is unclear. Maybe explain in comment?

apps/ui/src/components/pages/manage/utils/manageUtils.ts
JamesTLopez: Hmmm, I see what you mean here. Should we remove this `forApi` flag check, and then just increment by 1 just before antd displays the information?

demariadaniel (approved):
Tested on my end, migration works, tests pass, search bar works and from what I can see feedback has been addressed. Approving to unblock but lmk if I'm missing anything.

Issue #46, opened 2024-11-14
Add Revision Notes to UI


Buwujiu:
## Current State
When Rep and DAC request revisions, the notes are sent to applicants via emails, we need a better way to show the notes on the UI, so it's easier for applicants to make revisions without having to go back to the emails. 

## Functional Spec
https://docs.google.com/document/d/1zwxH6YnH2PQeekI2relFu1Obcf46zLFqOExbl8uqIr8/edit?tab=t.0

patrickdsantos:
Mocks are available here: https://www.figma.com/design/L4TWNoqAOOV9d0IBVUibIp/PCGL-Design-Mockups?node-id=5840-22130&t=LPGR27tAUv4aEY9v-1

Below is a preview of one screen (Applicant View)

<img width="1440" height="2267" alt="Image" src="https://github.com/user-attachments/assets/dcfb70a1-6255-4dbd-bd5e-668af3d0a899" />

@Buwujiu

Buwujiu:
Mockup looks great! thanks @patrickdsantos !

patrickdsantos:
Great work team! Can we reduce the padding for the body copy to 10px? See screenshot for reference.

<img width="939" height="403" alt="Image" src="https://github.com/user-attachments/assets/f6b04fc3-10a6-42cb-aea2-bb450a4ffcdc" />

Thank you! @Buwujiu @JamesTLopez

PR #491, opened 2025-11-03
#46: Revision Notes Feedback 

JamesTLopez:
## Summary

Full-stack implementation of the Display Revision Notes Feedback to each section.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/46

### API 

- Create new column in revisions table `isDacRequest`
   - To differentiate between Rep requests and Dac requests 
- Add `True and False` to `is_dac_request` variables in backend routes in DAC and Rep revision endpoints

### UI

- Refactor `useGetApplicationFeedback` hook
   - Transform revision object to accomodate feedback history from only most recent
   - Update components that used SectionRevision to accomodate Package changes
- Add RevisionsAlert component
- Implement RevisionsAlert component in all sections 

### Package
- fixed comment type from `String` to `string`
- add `isDacRequest` type
- add zod validation for `isDacRequest` in revisionDataResponse 
- update `SectionRevision` type
  - extend `GeneralType`
  - add `isDacRequest` and `createdAt` type in `RevisionType` 

### Special Instructions
Before running these changes, you will need to install  re-install and build the packages folder:
```
pnpm i && pnpm build:all
```

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/ui/src/api/queries/useGetApplicationFeedback.ts
```diff
@@ -27,68 +27,129 @@ import { SectionRevision } from '@pcgl-daco/validation';
...
-				return result;
+				return formatRevisionFeedback(data);
 			});
 		},
 	});
 };
 
 export default useGetApplicationFeedback;
+
+const formatRevisionFeedback = (data: RevisionsDTO[]): SectionRevision => {
```
JamesTLopez: Debating if we would benefit doing this in the backend instead?

Open to suggestions

apps/ui/src/components/RevisionsAlert.tsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+const RevisionsAlert = ({ sectionRevisions, general }: RevisionsProps) => {
+	const { token } = theme.useToken();
+
+	const { t: translate } = useTranslation();
+	const [filterRevisions, setFilteredRevisions] = useState<CollapseProps['items'] | undefined>(undefined);
+	const [generalComments, setGeneralComments] = useState<CollapseProps['items'] | undefined>(undefined);
+
+	useEffect(() => {
```
joneubank: This is a long useEffect, could you add some comments at the top to explain what is being done in this effect and when it is meant to trigger?

Can some of this be separated out into functions that are defined outside the useEffect?

joneubank (changes requested):
I want to revisit the database change, I think we should be getting our revision notes via joins with the application_actions table and then we would know if they were DAC or institutional rep review through the application action type. Send me a message if I'm not clear with this.

Issue #472, opened 2025-09-24
4.1 - Submit Comments - Design DB


Buwujiu:
## Descriptions
All DAC users can submit comments using message box inside the application page. 
Please read the functional spec for types of comments and rules. 

The first task is to design how DB should store comments.

PR #494, opened 2025-11-09
#472: DAC Comments DB Changes

JamesTLopez:
## Summary

> ~NOTE: These changes do not have generated migrations from drizzle-kit. Just the db-design, will generate the files if approved.~

DB changes to add DAC comments feature.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/472

## Thoughts

Comments do not work like revisions. Revisions are sent as a bulk, whereas comments can be sent individually per section, also multiple different DAC's can make a comment. 
As such, we need to know which comment belongs to which sections, hence why theres a section enum value.

Not 100% sold on the enum for the section, as there's the [shared packages enums](https://github.com/Pan-Canadian-Genome-Library/daco/blob/c53e768de4575376e9a0dd67996aa3ee3160e809/packages/validation/src/section/sections.ts#L20) we would have two points to maintain if we decided to add more sections in the future. I elaborate little bit in the review.

If there's a better design, open to suggestions or thoughts 👍 

## Description of Changes

- ADD comments table
- ADD section enums
- ADD new data-model.png
- ADD Drizzle related schema changes

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/db/schemas/comments.ts
```diff
@@ -0,0 +1,52 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { relations } from 'drizzle-orm';
+import { bigint, boolean, pgEnum, pgTable, text, timestamp, varchar } from 'drizzle-orm/pg-core';
+import { applications } from './applications.ts';
+
+export const sectionsEnum = pgEnum('sections', [
```
JamesTLopez: Many DAC members can leave comments on different sections, needed a way to differentiate which comments belonged to each section. 

Debated on adding this enum as now theres an already defined `SectionRoutes` in the
`packages/validation/src/section/sections.ts` shared package. So there will be two points of maintenance. 
Alternative could remove this enum, put it as a text, and populate with `SectionRoutes` programmaticly. We would have zod validation in the endpoints. Thoughts?

docs/model/data-model.dbml
```diff
@@ -204,4 +217,14 @@ table files {
   submitted_at timestamp [not null]
   content bytea [not null]
   filename varchar(255)
-}
\ No newline at end of file
+}
+
+Table comments {
```
JamesTLopez: Will probably use name `dac_comments`, as a better name. Open to suggestions.

docs/model/data-model.dbml
```diff
@@ -204,4 +217,14 @@ table files {
...
+  id bigint [pk, not null]
+  application_id bigint [not null, ref: > applications.id]
+  comments text
+  username text
```
joneubank: We need to capture the proper `user_id` which can be traced back to the user registration, but we also need to record the display name... so for consistency lets rename this to `user_name` and this should be `not null`.

docs/model/data-model.dbml
```diff
@@ -204,4 +217,14 @@ table files {
...
+  dac_chair_only boolean
+  section section [not null]
+  created_at timestamp [not null, default: `now()`]
+}
```
joneubank: Looks pretty good, just need to add `user_id` as a `[not null]` string field.

apps/api/src/db/schemas/comments.ts
joneubank: I think enum is good here. This is one of those "hard coded to our data model" decisions that simplifies so many other parts of the application but makes changing the application model and UI harder down the line. You can see we have a lot of things hardcoded to these sections, such as the text properties in the revision request table. This can always be shifted to something more flexible like text in the future if things change.

joneubank (commented):
Really good start, one addition and some renaming requested. Good to build on after that.

apps/api/src/db/schemas/dacComments.ts
```diff
@@ -0,0 +1,52 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { relations } from 'drizzle-orm';
+import { bigint, boolean, pgEnum, pgTable, text, timestamp, varchar } from 'drizzle-orm/pg-core';
+import { applications } from './applications.ts';
+
+export const sectionsEnum = pgEnum('sections', [
```
JamesTLopez: Was able to use `SectionRoutes` to populate the enums in postgres, this should alleviate possible mismatch routes in the future.

Issue #464, opened 2025-08-06
BE - Application History


Buwujiu:
## Tasks
1. There is an enum application_action_types is tracking all actions:
```
enum application_action_types {
  WITHDRAW
  CLOSE
  SUBMIT_DRAFT
  INSTITUTIONAL_REP_REVISION_REQUEST
  INSTITUTIONAL_REP_SUBMIT
  INSTITUTIONAL_REP_APPROVED
  DAC_REVIEW_REVISION_REQUEST
  DAC_REVIEW_SUBMIT
  DAC_REVIEW_APPROVED
  DAC_REVIEW_REJECTED
  REVOKE
}
```
Check if this is implemented correctly, if not, fix it. 
2. If the action types is implemented correctly, build a new api to generate application history.
3. Connect the UI with the application history endpoint.

PR #515, opened 2025-12-04
🪪 Feat/ Add Action User details

demariadaniel:
## Summary
Adds User Name to all Actions with related updates for Auth handling

### Related Issues
- [Application History BE](https://github.com/Pan-Canadian-Genome-Library/daco/issues/464)

## Description of Changes
- Adds standard handling for UserName to application Actions:
  - userName field for Actions
  - userName arg for state transitions
  - getUserName util function for standard userName def
- Adds Auth Type Guards & Request Handler
  - Adds isAuthenticatedRequest type guard & canAccessApplication in AuthService
  - Adds isAuthenticated request handler utility
  - Replaces all Router Auth checks & response handlers with above functions
- All Tests updated & passing

### Special Instructions
Requires running a database migration
`cd apps/api; pnpm drizzle-kit migrate`

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### DIFF
--- a/apps/api/src/routes/authRouter.ts
+++ b/apps/api/src/routes/authRouter.ts
@@ -225,7 +225,7 @@ authRouter.get('/user', async (request, response: ResponseWithData<UserResponse,
 	const { user } = request.session;
 
 	const output: UserResponse = {
-		role: getUserRole(request.session),
+		role: getUserRole(user),
 		user,
 	};
 


apps/api/src/service/utils.ts
```diff
@@ -43,3 +54,74 @@ export const applicationActionsQuery = (sort?: Array<OrderBy<ApplicationActionsC
...
+ */
+export function getUserName(user: SessionUser): string {
+	const { givenName, familyName, userId } = user;
+	const userName = givenName || familyName ? `${givenName || ''} ${familyName || ''}` : userId;
+	return userName.trim();
+}
+
+/**
+ * Standardized handler for common Auth error cases
+ * Insures Type safety for downstream Request Handler due to limitations with Express Type definitions
```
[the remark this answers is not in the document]
joneubank: Since we're proofreading, we also don't need to capitalize `Type`

apps/api/src/routes/signatureRouter.ts
```diff
@@ -234,71 +184,50 @@ signatureRouter.delete(
...
-							case 'SYSTEM_ERROR': {
-								response.status(500).json({ error: applicationResult.error, message: applicationResult.message });
-								return;
-							}
-							case 'NOT_FOUND': {
-								response.status(404).json({ error: applicationResult.error, message: applicationResult.message });
+				if (isAuthenticatedRequest(request) && (await canAccessRequest(request))) {
+					try {
+						// Validate Parameter
+						const applicationId = Number(request.params.applicationId);
```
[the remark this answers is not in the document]
joneubank: There is a validation wrapper around this request handler, unfortunately our method doesn't pass through the validated parameters atm so this pattern of converting the known safe value is being used. This is fragile to changes to the param type validation but isn't a new pattern here.

apps/api/src/middleware/authMiddleware.ts
```diff
@@ -55,21 +55,22 @@ type AuthenticationErrorResponse = ErrorResponse<['FORBIDDEN', 'UNAUTHORIZED']>;
  */
 export const authMiddleware =
 	(config: AuthMiddlewareConfig = {}): RequestHandler =>
-	(request, response: Response<AuthenticationErrorResponse>, next) => {
-		const { requiredRoles } = config;
+	async (request, response: Response<AuthenticationErrorResponse>, next) => {
```
joneubank: Can we remove this `async`?

apps/api/src/middleware/authMiddleware.ts
```diff
@@ -55,21 +55,22 @@ type AuthenticationErrorResponse = ErrorResponse<['FORBIDDEN', 'UNAUTHORIZED']>;
...
 				message: 'This resource is protected and requires authorization.',
 			});
+
 			return;
 		}
 
 		if (requiredRoles) {
-			const role = getUserRole(request.session);
-			if (!requiredRoles.includes(role)) {
+			if (!requiredRoles.includes(userRole)) {
```
joneubank: Looks like we can restore this entire function to how it originally was? We don't need to get the role until after we know there are required roles.

apps/api/src/service/authService.ts
```diff
@@ -76,3 +82,46 @@ export async function isAssociatedRep(session: Partial<SessionData>, application
...
+
+/**
+ * Handler to enforce Request is Authorized
+ * Insures Type safety for downstream Request Handler due to limitations with Express Type definitions
+ * @param request Express request
+ * @returns boolean
+ */
+export const isAuthenticatedRequest = (request: Request): request is AuthorizedRequest => {
+	return isRequestWithSession(request);
+};
```
joneubank: We don't need this. Duplicate function with new name.

apps/api/src/service/authService.ts
```diff
@@ -76,3 +82,46 @@ export async function isAssociatedRep(session: Partial<SessionData>, application
...
+ * @param request AuthorizedRequest with session, user and userId
+ * @returns boolean
+ */
+export async function canAccessRequest(
+	request: AuthorizedRequest,
+): AsyncResult<AuthorizedRequest, 'FORBIDDEN' | 'NOT_FOUND' | 'SYSTEM_ERROR'> {
+	const { session } = request;
+	const { user } = session;
+	const userRole = getUserRole(session);
+	const requestedId = request.params.applicationId || request.body.applicationId;
```
joneubank: Make this a function parameter `applicationId`. We discussed not hardcoding information about the request structure into this function. We should just take the ID as a parameter, and let the original handler pull out the application ID from wherever it is.

This would also let us reuse this logic for an array of different applications.

apps/api/src/service/authService.ts
```diff
@@ -76,3 +82,46 @@ export async function isAssociatedRep(session: Partial<SessionData>, application
...
+};
+
+/**
+ * Validate User is allowed access to this specific Application based on userRole or userId
+ * @param request AuthorizedRequest with session, user and userId
+ * @returns boolean
+ */
+export async function canAccessRequest(
+	request: AuthorizedRequest,
+): AsyncResult<AuthorizedRequest, 'FORBIDDEN' | 'NOT_FOUND' | 'SYSTEM_ERROR'> {
```
joneubank: Could we refactor this to not take the `request` object, but instead take the `session` and `applicationId` as parameters?

apps/api/src/service/authService.ts
```diff
@@ -76,3 +82,46 @@ export async function isAssociatedRep(session: Partial<SessionData>, application
...
+		const { userId } = user;
+		const canAccess = data.userId === userId || hasSpecialAccess;
+
+		if (!canAccess) {
+			return failure('FORBIDDEN', 'User does not have permission to access or modify this application.');
+		}
+	} else {
+		return result;
+	}
+	return success(request);
```
joneubank: I think we can just return a `Result<void>` since there is no new information returned on success, only error codes on failure. This is the equivalent of returning "success OR some error code".

apps/api/src/service/utils.ts
```diff
@@ -43,3 +54,74 @@ export const applicationActionsQuery = (sort?: Array<OrderBy<ApplicationActionsC
...
 };
+
+/**
+ * Type Guard to validate session contains user data with userId
+ * @param session express-session session object / Express Request['session']
+ * @returns boolean
+ */
+export const isSessionWithUser = (session: SessionType): session is UserSession => {
+	return typeof session.user !== 'undefined' && !!session.user.userId;
+};
```
joneubank: I would be cautious creating too many explicit type predicates. Type predicates are very similar to type assertions, in that the TS compiler is trusting that you did all the type checking correctly.

For example, with this function, if the input type (SessionType) changes such that `session.user` is no longer `UserSession`, then this type narrowing is invalid.

It is almost always best to let TypeScript and ONLY TypeScript perform type narrowing. TS since 5.5 is capable of inferring the type predicate of a function that has no control logic (switch/if/else) and returns a boolean statement. There is something weird with this specific type that is preventing this from working in this specific case, and while I want to understand why, solving it is taking longer than the value that would be gained.

My recommendation: Remove this utility and utility type. It is just as easy to do `session.user &&` as it is to do `isSessionWithUser(session) &&`, and in many cases will be more clear to the reader what check is being done.

I would repeat this recommendation for the other type predicates you have introduced.

joneubank (changes requested):

apps/api/src/service/authService.ts
demariadaniel: Sure, crossed my mind too. Thought there may be value in case we wanted to expand the request handler version.

apps/api/src/service/utils.ts
demariadaniel: Discussed in office, replaced use of type guard with basic boolean narrowing

```
const { user } = request.session;
if (!user) { ... } else { authErrorHandler; }
```

Issue #523, opened 2025-12-10
Multiple DAC - User Session Updates


joneubank:
## Description

The User session needs to be modified due to the introduction of DAC Chairs for Local DACs in addition to the PCGL DACO Members and Chairs.

This ticket will modify the session object to track which DACs a user is a chair of or member for.

Due to the change in the session structure, we will need to update our Auth middleware and other logic throughout the API that uses the old User Role system. Separate tickets will be opened to track those specific rules changes.

### Notes about Roles
In practice, a user will only be the Chair or Member of a single DAC, and Local DACs will only have Chairs and no members. Despite this, we can easily build this model to support users assigned to multiple roles. More importantly, there will be weird edge cases that can't be well handled if we try to code logic for users to have a single DAC membership, since this is managed through co-manage groups where they could have multiple membership.

### DAC Membership Groups
- We will now have multiple DACs to support.
- Each DAC has a unique ID.
- A user can be a Member and/or a Chair of any number of DACs. 
- DAC Membership is based on CoManage groups, the list of group memberships is provided by the AuthZ response, as it is already, but now the group names will be templated using each DAC's `daco_id`, so there could be any number of DAC Chair and DAC Member groups.

- [x] The group name template for a DAC chair is `PCGL:DACO:DAC-Chair:${daco_id}`
- [x] The group name template for a DAC member is `PCGL:DACO:DAC-Member:${daco_id}`

### User Session Structure
- [x] We can remove the user ROLE from the session, our existing Roles no longer work with the multiple DAC model.
- [x] We need to add two new properties to the session
   ```ts
   {
     dacChair: string[]; // Contains the daco_id values for every DAC they are the Chair of
     dacMember: string[]; // Contains the daco_id values for every DAC they are a Member of
   }
   ```

### UserResponse schema for `/user` Auth route
- [x] UserResponse schema updated to include list of DAC membership properties from session
- [x] /user request returns updated session info

## Breaking Changes
Modifying the user session is a change that will break auth logic on the API and UI. We should make this change as part of a feature branch for Multiple DAC Auth so that we can review it separately but also combine it with other auth updates before merging into our main branch.

Issue #386, opened 2025-05-01
Multiple DACs


Buwujiu:
## Description
User Story: https://docs.google.com/spreadsheets/d/13h-jKeeEtmEV5PvW8pTgWyjPepI60HXV2yVkOQZI7U8/edit?gid=1966145028#gid=1966145028

Functional Spec: https://docs.google.com/document/d/1zwxH6YnH2PQeekI2relFu1Obcf46zLFqOExbl8uqIr8/edit?tab=t.0#heading=h.c4u1hzmofglj

This epic details the steps to add support for multiple DAC organizations to share the PCGL DACO system. There are two types of DAC:
1. The PCGL DAC - operated by the PCGL DACO and manages most studies
2. Local DAC - operated by DACs, but using the PCGL DACO UI and API to manage applications to a subset of "assigned studies" managed by that DAC.

As detailed in the user stories, an individual application can select multiple studies, but all studies must be managed by the same DAC.

The DACO API will import DACs and studies that are registered in the PCGL Clinical Submission system. This is the data that will be used to determine which DAC will be responsible for approving requests to access each study. This data will be replicated into the DACO database, but will be overwritten when updates are fetched from the Clinical Submission system. Studies are not automatically available to be applied to after being imported from Clinical Submission; a DACO Admin must activate a study using the API to make it available to create applications for access to that study.

The user for the chair of each Local DAC will be assigned through CoManage using a group with name templated for the DAC. We will also update the groups that identify PCGL DACO members and chair to match this pattern. Members of a DAC (Local or PCGL) will only be able to see and review applications that are submitted for the DAC which that user represents.

Applicants will be able to apply to multiple studies, but all studies must belong to the same DAC. After selecting their first study in the UI, studies belonging to other DACs should not be selectable (disabled options or filtered out, pending UX decision).

## Tasks

- DACO API
	- [ ] #523 
	- DB Model Updates
		- [ ] #525
		- [ ] https://github.com/Pan-Canadian-Genome-Library/daco/issues/529
	- Submission Service
		- [ ] Update Study Data https://github.com/Pan-Canadian-Genome-Library/Roadmap/issues/148
			- Fetches study from submission service, and fetches the DAC for that study by study ID, then it will upsert both records.
		- [ ] https://github.com/Pan-Canadian-Genome-Library/Roadmap/issues/149
	- Routes
		- [ ] https://github.com/Pan-Canadian-Genome-Library/daco/issues/530
[...]

PR #568, opened 2026-02-11
#523: Multi-DAC Session Updates 

JamesTLopez:
> NOTE: These changes break frontend as the role has been removed, will be done in separate tickets

## Summary

Update session on the UI and API. Return select fields to the UI and add additional fields `dacChair` and `dacMember`.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/386
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/523

### API

- Add `dacChair` to session
   -  Filter through group to check if prefix matches `AUTHZ_GROUP_DAC_CHAIR`
   -  Map through filtered values to return only `dac_id`
- Add `dacMember` to session
   -  Filter through group to check if prefix matches `AUTHZ_GROUP_DAC_MEMBER`
   -  Map through filtered values to return only `dac_id`
- Update `GET /user` endpoint
   - Return only values defined from `{ user: SessionUserUI | undefined }` 

### Packages

- Create type `SessionUserUI`
- Create zod validation for `dacChair`
- Create zod validation for `dacMember`

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### DIFF
--- a/packages/validation/src/routes/auth.ts
+++ b/packages/validation/src/routes/auth.ts
@@ -125,7 +125,7 @@ export const sessionUser = authGeneratedSessionValues.merge(dacoGeneratedSession
 
 export type SessionUser = z.infer<typeof sessionUser>;
 
-export const SessionUserResponse = sessionUser.pick({
+export const sessionUserResponse = sessionUser.pick({
 	userId: true,
 	givenName: true,
 	familyName: true,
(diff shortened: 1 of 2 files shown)


packages/validation/src/routes/auth.ts
```diff
@@ -121,14 +121,21 @@ const authGeneratedSessionValues = z.object({
...
+	userId: true,
+	givenName: true,
+	familyName: true,
+	emails: true,
+	dacChair: true,
+	dacMember: true,
+	dacoAdmin: true,
+});
+
+export type SessionUserResponseType = z.infer<typeof SessionUserResponse>;
```
joneubank: Types do not need to be named `Type`, we know its a type because of what it is. To help differentiate type from the zod schema, the schema can be camel cased, or name it with schema suffix. In this app it looks like schemas are named camel case (see `sessionUser` above).

Issue #569, opened 2026-02-12
Modify List Applications to Filter by Dac Id


demariadaniel:
## Descriptions
Modify existing route - List applications
When called by a DAC chair, this needs to filter applications by the DAC ID that they are the chair of (could be chair of multiple...)
When called by a DAC member, needs to filter by the DAC ID of the PCGL DAC (from config)

Follow up from: https://github.com/orgs/Pan-Canadian-Genome-Library/projects/1/views/1?pane=issue&itemId=147992617&issue=Pan-Canadian-Genome-Library%7Cdaco%7C532

Prerequisite: https://github.com/Pan-Canadian-Genome-Library/daco/issues/523

PR #576, opened 2026-02-26
#569 Feat / Filter List Applications by DAC :receipt:

demariadaniel:
## Summary
Updates Get `/applications` endpoint to return applications belonging to DAC groups where User has permissions

### Related Issues
- #569

## Description of Changes
- `GET /applications/` updated to filter Application records based off DAC ID
- Includes Applications where userId is a match, plus `application.dacId` is found in `user.dacMember` or `user.dacChair`
  
Screenshot with no Dac permissions (6 total matching UserId):
Screenshot 2026-03-05 at 12 03 22 PM

Screenshot w/ Dac permissions (9 total matching UserId & Dac permissions):
Screenshot 2026-03-05 at 12 03 46 PM

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### DIFF
--- a/apps/api/tests/services/application-service.test.ts
+++ b/apps/api/tests/services/application-service.test.ts
@@ -374,7 +374,6 @@ describe('Application Service', () => {
 					institution_rep_approved: false,
 					collaborators_approved: false,
 					project_approved: false,
-					requested_studies_approved: false,
 				},
 			});
 


apps/api/src/controllers/applicationController.ts
```diff
@@ -171,9 +171,10 @@ export const editApplication = async ({
  * @param sort - sorting options
  * @param page - page offset
  * @param pageSize - page limit
- * @param isDACMember - Boolean which represents if the user is a DAC Member (they can see all applications)
+ * @param isDAC - Boolean which represents if the user is a DAC Member (they can see all applications)
```
joneubank: Remove this parameter, no longer usd.

apps/api/tests/api/application-controller.test.ts
```diff
@@ -136,6 +137,24 @@ describe('Application API', () => {
 		});
 	});
 
+	describe('Get All Applications', () => {
+		it('should retrieve applications with submitted user_id and no DAC IDs', async () => {
+			const testResult = await getAllApplications({ userId: user_id });
+
+			assert.ok(testResult.success && testResult.data);
+			assert.ok(testResult.data.applications.every((record) => record.userId === user_id));
+			assert.ok(testResult.data.applications.every((record) => record.dacId === null));
```
joneubank: I don't see from this test what the test data is... but it is entirely expected that you will be able to get applications that are already assigned to a DAC. This statement COULD be true for some data sets, but is not true in general.

What you should be testing here is:
1. all applications returned belong to this user (tested correctly)
2. that we have the full count of all applications that belong to this user from the test data set

joneubank (changes requested):
Implementation looks correct, but the test cases need some updates.

apps/api/src/service/applicationService.ts
```diff
@@ -353,17 +360,21 @@ const applicationSvc = (db: PostgresDb) => ({
...
 				.where(
 					and(
-						user_id ? eq(applications.user_id, String(user_id)) : undefined,
 						search ? transformSearchIntoQuery(search) : undefined,
+						// Retrieves Applications where the current User is the Applicant,
+						// and Applications where the User is a DAC Member or Chair
+						or(
+							user_id ? eq(applications.user_id, String(user_id)) : undefined,
+							authorizedDacIds?.length ? or(inArray(applications.dac_id, authorizedDacIds)) : undefined,
+						),
```
JamesTLopez: The logic above should work for this as well.

Issue #625, opened 2026-04-30
3 - Update Study API Services


JamesTLopez:
### Description

Update `studyController` and `studyService` to match database changes with `study_translations` and dropped columns from `study`

Issue #619, opened 2026-04-30
Study Localization DACO


JamesTLopez:
### Description

DACO fetch `study` records from clinical-submission to populate the `requested studies` section in the daco application process. Clinical submission recently has changes that altered the table as well as added an additional  
table `study_translations`. We need to update daco inherited table `study` to the new changes as well as add `study_translation` table.

Refer to below for clinical-submission:
- https://github.com/Pan-Canadian-Genome-Library/clinical-submission/issues/148

Buwujiu:
Tested DACO staging relevant functions, working as expected!

PR #631, opened 2026-05-08
#625: Create Study Upsert Service

JamesTLopez:
## Summary

Study localization study upsert for importing studies from clinical and also create just translations for studies.
These services are not endpoints accessible to daco, only used for syncing the study and study_translation tables.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/625
- https://github.com/Pan-Canadian-Genome-Library/daco/issues/619

### API

- add `upsertStudy` controller
- add `createStudyFromClinical` service, creates a CTE to insert study and study translation at the same time
- add `createStudyTranslation`service, creates specific translation for a study
- update and add upsert types for `createStudyFromClinical` parameters

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


JamesTLopez:
Found an issue when writing the test cases, back into draft.

JamesTLopez:
> Found an issue when writing the test cases, back into draft.

Resolved ready for review

apps/api/src/controllers/studyController.ts
```diff
@@ -90,37 +89,34 @@ export const getAllStudies = async (): AsyncResult<StudyDacoDTO[], 'SYSTEM_ERROR
...
 		const database = getDbInstance();
 		const studyService = studySvc(database);
 
-		const studyData = studies.map((study) => {
-			const createdDate = typeof study.createdAt === 'string' ? new Date(study.createdAt) : study.createdAt;
-			const updatedDate = typeof study.updatedAt === 'string' ? new Date(study.updatedAt) : study.updatedAt;
+		for (const study of studies) {
+			const studyModel: UpsertStudy = {
+				...study,
+				defaultLanguage: study.translations[0]?.languageId || 'en_ca',
```
leoraba: taking the index 0 languageId won't always be the default language, however clinical doesn't seem to return the default language code in the response so this seems like something we should take care later, can you add a comment about this?

apps/api/src/controllers/studyController.ts
```diff
@@ -90,37 +89,34 @@ export const getAllStudies = async (): AsyncResult<StudyDacoDTO[], 'SYSTEM_ERROR
...
+				createdAt: typeof study.createdAt === 'string' ? new Date(study.createdAt) : study.createdAt,
+				updatedAt: typeof study.updatedAt === 'string' ? new Date(study.updatedAt) : study.updatedAt,
+				acceptingApplications: false,
```
leoraba: `acceptingApplications` is not being used in the `createStudyFromClinical` function, maybe we should remove it

apps/api/src/routes/studyRouter.ts
leoraba, line 1: should the GET endpoints also return the `translations` for each study?

leoraba (dismissed):
Tested locally and works as expected 👍 , just left small comments

apps/api/src/controllers/studyController.ts
JamesTLopez: I don't think this is also needed, this was a param for clinical-submission when you use the `POST /study/` as you would need to explicitly say what language you are submitting in. Since we are not actually creating studies in DACO, we can remove this along with `acceptingApplications `(since acceptingApplications defaults to false anyways)

apps/api/src/routes/studyRouter.ts
JamesTLopez: I think we can hold off on this for now for two reasons:

1. We do not utilize any of the fields in `translations` in DACO atm
2. As a result, we would do more db calls for all the study records to append `translations` for something we wouldn't use.

Open to suggestions

apps/api/src/controllers/studyController.ts
leoraba: that's even better, removing unnecessary fields and the type `UpsertStudy`. thanks.

leoraba (approved):

Issue #659, opened 2026-06-04
Record Revoke and Reject Application Reasoning in Database


JamesTLopez:
## Descriptions

Currently, when a DAC chair revokes or rejects an application and provides a reasoning, there is no record of the comment sent being saved in the database.

## Acceptance Criteria

- Add comment column in emails table
- Adjust revoke and reject email services to pass the reasoning to the email record

PR #660, opened 2026-06-04
#659: Record Revoke and Reject Email Reasoning In Emails Table

JamesTLopez:
## Summary

Generate migration to add `comment` property to emails table to record revoke and reject email reasoning.

### Related Issues

- https://github.com/Pan-Canadian-Genome-Library/daco/issues/659

### API

- Generate db migration to add `comment`
- Update Revoke and Reject email services to pass reasoning to new `comment` field record

## Readiness Checklist

- [x] **Self Review**
- [x] **PR Format**
- [x] **Labels Added**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation


apps/api/src/service/email/emailsService.ts
```diff
@@ -893,6 +896,7 @@ const emailSvc = (db: PostgresDb) => {
 					application_action_id: actionId,
 					email_type: EmailTypes.NOTIFY_APPLICANT_DAC_REVIEW_REJECTED,
 					recipient_emails: [to],
+					comment,
```
leoraba, line 875: Question: Do we also need to add the `comment` when application gets `revoked`?
what about other events like revisions, do we need to store those comments?

JamesTLopez: From last time I discussed with Jon. It was only required to do track reasoning.

JamesTLopez: We also already have a revisions table that records revisions

leoraba: thanks for clarifying, in that case it's also needed to add `comment` when application gets `revoked` and `closed` if there's a comment variable for those.

joneubank: Can't remember the decision last time we spoke James, but I agree with Leo that it would be best to have comment on the revoke case as well. Revisions have a different mechanism for comments, but these two cases should allow the admin to explain **why** they took the action.

leoraba (approved):
LGTM

